PR #4, opened 2017-10-13
extensive changes to support DALI-1.1 datatypes

pdowler:


cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/votable/VOTableUtil.java
```diff
@@ -0,0 +1,116 @@
...
+public abstract class VOTableUtil 
+{
+    private static final Logger log = Logger.getLogger(VOTableUtil.class);
+
+    private VOTableUtil() { }
+    
+    public static int[] getArrayShape(String arraysize)
+    {
+        if (arraysize == null || arraysize.equals("1")) // interpretation TBD
+            return null;
```
andamian: Don't you want to throw an exception here? Or null array shape is a valid case?

cadc-dali/src/main/java/ca/nrc/cadc/dali/util/DoubleArray2DFormat.java
```diff
@@ -0,0 +1,169 @@
...
+        
+        StringBuilder sb = new StringBuilder();
+        for (int i=0; i<object.length; i++)
+            for (int j=0; j<object[i].length; j++)
+            {
+                sb.append(Double.toString(object[i][j]));
+                sb.append(" ");
+            }
+        return sb.substring(0, sb.length() - 1); // trim trailing comma
+    }
```
andamian: trailing space hence a trim() would be more suitable maybe.

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/votable/VOTableUtil.java
pdowler: null arrayShape means it is a scalar value

cadc-dali/src/main/java/ca/nrc/cadc/dali/util/DoubleArray2DFormat.java
pdowler: I changed a bunch of array formatters to use StringBuilder substring to trimToSize and toString; the separator used to be comma hence the old trim hack :-)

PR #13, opened 2019-04-15
Fixes #12

### LINKED ISSUE #12: Make TAP URL configurable in cadc-sia
There currently exists an `SiaRunner.properties` file to configure a TAP Resource ID to use.  This is useful for registered TAP services.

However, we could also allow absolute URLs for unregistered TAP Services.  We could also re-use the existing `tapURI` property.

at88mph:


cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -130,27 +146,60 @@ public void setState(String string)
...
+        if (configuredTapURI.getScheme().startsWith("http"))
+        {
+            return configuredTapURI.toURL();
+        }
+        else
+        {
+            AuthMethod am = AuthenticationUtil.getAuthMethod(AuthenticationUtil.getCurrentSubject());
+            return regClient.getServiceURL(configuredTapURI, Standards.TAP_10, (am == null) ? AuthMethod.ANON : am);
+        }
+    }
```
brianmajor: I wonder if the order of operations should be switched here... first check if the reg client returns a URL.  If not, just return uri.toURL().  That would avoid the 'if http' check.
You'd have to catch the MalformedURLException on the toURL() call and return null.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -86,29 +88,43 @@
  *
  * @author Sailor Zhang
  */
-public class ServiceAvailability implements WebService
+public class ServiceAvailability implements AvailabilityPlugin
 {
     private static final Logger log = Logger.getLogger(ServiceAvailability.class);
-    
+    private static final String DEFAULT_TAP_URI = "ivo://cadc.nrc.ca/tap";
```
brianmajor: I don't think there should be the concept of a default URI in the code.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -130,27 +146,60 @@ public void setState(String string)
...
-            String fname = "SiaRunner.properties";
+            final String configFileLocation = System.getProperty("user.home") + "/config/SiaRunner.properties";
             try
             {
-                URL url = SiaRunner.class.getResource(fname);
-                if (url == null)
-                    url = SiaRunner.class.getResource("/" + fname);
-                Properties props = new Properties();
+                final URL url = new URL("file://" + configFileLocation);
+                final Properties props = new Properties();
```
brianmajor: Why change the lookup from classpath to file location?

brianmajor: (If we go this route, there's a PropertiesReader file that will do this work for you)

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
at88mph: Should it not be usable out of the box, though?  Should the default be put into the configuration, then?

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
at88mph: To make it externally configurable.  This PR is meant to support ALMA, who pushes their releases through QA which uses new release hostnames each time.  It would be much easier to have it externally configurable rather than rebuilding the Docker Image each time.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
at88mph: Done.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
pdowler: out sia service needs the config file like everyone else is OK

at88mph: Changed to check the class path first, then the external file system based on the `PropertiesReader` looks for external configuration.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
brianmajor: try/catch around configuredTapURI.toURL() to catch MalformedURLException so that null can be returned...

at88mph:
Thanks for the reviews.  When it's published I can proceed with the ALMA build.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
brianmajor: Just one more comment Dustin:  Could you change the signature of this method so that it throws MalformedURLException (and potentially others) rather than blanket Exception?

PR #27, opened 2019-09-25
Fixes #26

### LINKED ISSUE #26: Prefer override config file over class path one.
The `ServiceAvailability` class prefers the class path version of the `SiaRunner.properties`, so it can't be overridden in the `${HOME}/config` unless it's removed from the class path in the JAR.

at88mph:


cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -155,17 +157,23 @@ public void setState(String string) {
...
-                (am == null) ? AuthMethod.ANON : am);
+        try {
 
-        return (serviceURL == null) ? configuredTapURI.toURL() : serviceURL;
+            // Attempt to load the URI as a resource URI from the Registry.
+            return regClient.getServiceURL(configuredTapURI, Standards.TAP_10,
+                                           (am == null) ? AuthMethod.ANON : am);
+        } catch (IllegalArgumentException iae) {
+            // Fallback and assume the URI is an absolute one.
+            return configuredTapURI.toURL();
```
brianmajor: IllegalArgumentException is only thrown when one of the params is null... you could check those in this code instead then you wouldn't be relying on the exception to make the decision.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -187,18 +195,20 @@ public static String getTapURI() {
...
-        if (url == null) {
-            url = SiaRunner.class.getResource("/" + CONFIG_FILE_NAME);
-        }
+        // Try to get from a file on disk first.
+        final PropertiesReader propertiesReader = new PropertiesReader(CONFIG_FILE_NAME);
 
-        if (url == null) {
-            final PropertiesReader propertiesReader = new PropertiesReader(CONFIG_FILE_NAME);
+        if (propertiesReader.canRead()) {
             final MultiValuedProperties multiValuedProperties = propertiesReader.getAllProperties();
```
brianmajor: PropertiesReader has a getFirstValue(key) method that would allow you to make these checks a little cleaner.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -187,18 +195,20 @@ public static String getTapURI() {
...
             final List<String> props = (multiValuedProperties == null)
-                    ? new ArrayList<String>() : multiValuedProperties.getProperty(CONFIG_TAP_URI_KEY);
+                                       ? new ArrayList<String>() : multiValuedProperties.getProperty(
+                    CONFIG_TAP_URI_KEY);
             return props.isEmpty() ? null : props.get(0);
         } else {
+            URL url = SiaRunner.class.getResource(CONFIG_FILE_NAME);
+            if (url == null) {
+                url = SiaRunner.class.getResource("/" + CONFIG_FILE_NAME);
+            }
```
brianmajor: Not so fond of the 'try both of these' approach... isn't there only one correct way to do it?  (with or without a slash?)

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
at88mph: The `IllegalArgumentException` is thrown from the `getAccessURL` method unfortunately, and so needs to be dealt with.  That may be a bug.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
at88mph: This exists for Java 8 and Java 11 support.  Java 9 (and later) requires the leading slash.

settings.gradle
```diff
@@ -0,0 +1,2 @@
+rootProject.name = 'dal'
+include 'cadc-dali', 'cadc-dali-pg', 'cadc-datalink-server', 'cadc-sia', 'cadc-soda-server'
```
pdowler: What is this settings.gradle for?

at88mph: This is used at the top level by Gradle.  If, while in the root of the project, I run `gradle -i clean build test`, it will iterate over the `include` projects and run those gradle tasks on them.  It good for making sure that I didn't break anything else in the repository when I make a change.  It may simplify the `.travis.yml` file, too.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
pdowler: There is a method in FileUtil (cadc-util) that does something like this... it should be used (and fixed if necessary) so that when we go to java 9+ and/or someone decides to change classloader semantics we only have one  bug to fix. Then the swugly is only in one place... a comment to explain that ugliness will help prevent it from being broken.

at88mph: OK.  Will change.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
pdowler: That IllegalArgumentException is at a minimum not thrown for the right reason. The reason it should be thrown is that the argument resourceID is not an ivo-scheme URI and it could thrown a ResourceNotFoundException or return null if the service is not "registered". 

The code here should not rely on that behaviour or use exception for logic. You should just check that the URI scheme is ivo before calling regClient and try toURL if it is not.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -155,17 +154,23 @@ public void setState(String string) {
...
         RegistryClient regClient = new RegistryClient();
         URI configuredTapURI = URI.create(ServiceAvailability.getTapURI());
 
         AuthMethod am = AuthenticationUtil.getAuthMethod(AuthenticationUtil.getCurrentSubject());
-        URL serviceURL = regClient.getServiceURL(configuredTapURI, Standards.TAP_10,
-                (am == null) ? AuthMethod.ANON : am);
+        try {
 
-        return (serviceURL == null) ? configuredTapURI.toURL() : serviceURL;
+            // Attempt to load the URI as a resource URI from the Registry.
```
pdowler: That IllegalArgumentException is at a minimum not thrown for the right reason. The reason it should be thrown is that the argument resourceID is not an ivo-scheme URI and it could thrown a ResourceNotFoundException or return null if the service is not "registered". 

The code here should not rely on that behaviour or use exception for logic. You should just check that the URI scheme is ivo before calling regClient and try toURL if it is not.

at88mph: Fixed now, but relies on `cadc-util:1.2.23` to be published.

cadc-sia/src/main/java/ca/nrc/cadc/sia2/impl/ServiceAvailability.java
```diff
@@ -149,23 +149,32 @@ public void setState(String string) {
         //no-op
     }
 
+    private static boolean isIVOURI(final URI uri) {
```
pdowler: there will never (> 99%) be a more elaborate check here so please inline this where it is used

at88mph: Done.

PR #38, opened 2020-11-04
cadc-data-fits-ops implementation

at88mph:
- Making use of NOM TAM FITS library.
 - Performs simple pixel cutout on MEF file (included in `src/test/resources`).
 - Will not compile as it relies on `nom-tam-fits:1.15.3-SNAPSHOT` with PR #146 (https://github.com/nom-tam-fits/nom-tam-fits/pull/146).


pdowler (commented):
Initial comments:

We have log4j 1.2 in almost all our other code - this uses 2.x - a major version change. What's going to happen when this lib is included with something that was built with 1.2? We could change to 2.x but that's a lot of work...

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/Range.java
```diff
@@ -0,0 +1,143 @@
...
+package org.opencadc.fits.slice;
+
+import java.util.Iterator;
+
+
+/**
+ * Represents a Range of integer values.  Can be expanded to an entire array, but only on demand.  This uses concepts
+ * from Python's numpy project to be able to have a start (lowerBound), stop (upperBound), and a striding value (step).
+ */
+public class Range {
```
pdowler, line 78: This class name could be more specific, maybe PixelRange?

The 3-arg ctor should call `this(lowerBounds, upperBound)` then just assign `step`.

That single-arg constructor seems like it will inevitably be ambiguous: if there is a use case for (,upper] then there is also a use case for [lower,) ... the correct solution is probably to use the two-arg ctor and change the type to Integer so that null values can be used for open range (and then carefully handle them once you know the actual dimensions). If you can just drop that 1-arg ctor then ignore what's below.

If you do need open-ended range, then it has to be Integer and null allowed. The best practice that we've been using in other code to good effect is to declare thing that can be null as public fields, so if Integer is the correct thing to use here then lowerBound and upperBound would be 
```
public final Integer lowerBound;
public final Integer upperBound;
```
(yeah, public final looks wierd but works as expected).

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/Slices.java
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Represents a single extension and set of pixel ranges.
+     */
+    public static final class ExtensionSliceValue {
+
+        private final static String ALL_DATA = "*";
+        private final static String SEPARATOR = ",";
+        private final static String PIXEL_VALUE_DELIMITER = ":";
+        private static final Pattern VALID_SINGLE_RANGE = Pattern.compile("[\\d*\\*]+:?\\d*");
```
pdowler, line 136: The pattern value here is a subset of PIXEL_RANGE_PATTERN_STRING so this should be defined above that and used when creating PIXEL_RANGE_PATTERN_STRING - because they do have to agree on this.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/FitsOperations.java
```diff
@@ -130,4 +131,22 @@ public Header getPrimaryHeader() throws ReadException {
             throw new ReadException("failed to read " + src.getAbsolutePath(), ex);
         }
     }
+
```
pdowler, line 134: Not sure about the method name or the `String cutoutSpec` arg, but also don't have anything better to suggest right now. I know that just passing String around ends up being a total non-type-safe mess eventually... We'll eventually want to separate parsing/validating string input from this method, but that will happen later on.

pdowler (changes requested):
Now that this lib depends on the forked nom-tam-fits with the RandomAccessDataObject interface, the FitsOperations class should accept that (only) as the input data (rather than File).

The local unit tests can use RandomAccessFileExt as the implementation.

at88mph:
Thank you.  Rework submitted.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java
```diff
@@ -0,0 +1,399 @@
...
+        } catch (FitsException fitsException) {
+            throw new IllegalStateException(fitsException.getMessage(), fitsException);
+        }
+
+        slice(fits, extensionSliceValues, output);
+    }
+
+    private void slice(final Fits fits, final Slices.ExtensionSliceValue[] extensionSliceValues,
+                       final ArrayDataOutput output) throws FitsException, IOException {
+        // Assume MEF if more than one extension was requested, so write a Primary HDU before anything else.
```
pdowler, line 159: This logic is not going to hold up and one cannot rely on the input requested slices to decide what to output.

I think of FITS files as having 3 broad types:
1. simple FITS file with data in the primary HDU
2. me FITS file with no data in the primary and 1+ extensions
3. me FITS file with data in the primary HDU and 1+ extensions
While 2 and 3 look the same structurally, there is one important distinction: in 2, extensions "inherent" metadata from the primary header -- in 3, no metadata is inherited by extensions and each is self contained... you could probably think about 3 as being "just extensions and no common metadata" and 1 could be thought a special case of 3.

So, the first thing is to determine if this is case 2 or not and that means reading the primary header to find out.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java
```diff
@@ -0,0 +1,399 @@
...
+            final BasicHDU<?> firstHDU = fits.getHDU(0);
+            final Header headerCopy = ImageHDU.manufactureHeader(firstHDU.getData());
+            final Header primaryHeader = firstHDU.getHeader();
+
+            copyHeader(primaryHeader, headerCopy);
+
+            final HeaderCard nextEndCard = headerCopy.findCard(DataDescription.NEXTEND);
+
+            // Adjust the NEXTEND appropriately.
+            if (nextEndCard == null) {
```
pdowler, line 170: This is going to be problematic. At this point we do not know how many extensions are going to be output so we do not know what value to set NEXTEND to. For example, the sample-mef.fits has a primary and 3 extensions: what if the caller asks for pixels from extension 7? This code would set NEXTEND = 1 but then not be able to output an extension; the output would be invalid. 

Yeah, fits has several things like this that make streaming output hard and this is one of them. The first thing is to verify what happens when NEXTEND is missing or wrong (both from a spec point of view and whther nom-tam-fits and fitsverify think it is OK or complain).

To be strictly correct, one would have to figure out how many extensions will actually be output, which means iterating through the HDUs until you find the ones that match the requested slices. It looks like the Fits object keeps the already-read HDUs in a list and that `getHDU(n)` re-uses that so you can jump back without re-reading. 

Note: It remains to be verified what IO ops occur when doing this and if going back and slicing data works as expected. It is clear that an MEF file with many extensions will potentially store a lot of header data in memory (something to worry about later). It's good to keep an eye on the IO ops and see when a change in code changes the IO ops being performed.

So, in addition to dealing with the 3 types of FITS files above, I would suggest a loop like the one in `FitsOperations.getHeaders()` that checks each header to determine if it is to be part of the output (and stopping loop as soon as all requested extensions are accounted for). After that point, you will know the right value for NEXTEND, will be able to output extensions in either original (file) order or requested (slice) order, and -- when we do get to applying sky coord cutouts -- we could have already done the intersection of the sky shape with the data array.

pdowler (changes requested):
Partial review of NDimensionalSlicer...

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java
at88mph: Does this suggest that we should write out the same style as the input file?

I can think of two use cases where that would not be appropriate:

1. [0][100:300][0][3000:4000] - Two cutouts from a simple FITS file.  Couldn't this be a possibility with a big cube like VLASS?  This sounds like an MEF.
1. [121][300:400] - A single cutout from an MEF.  This sounds like a simple FITS.

Finally, I'm sure there are cases where there is no overlap from any requested cutout, in which case you would get an MEF (if cutting from one) with just a primary header or (possibly) _just_ a header for each requested cutout.

pdowler: For the first case: yes, that should be an MEF result (1 ext per cutout). You may have to change SIMPLE and add NEXTEND keywords if the input was simple. In the case where you really mean extension 0 (so either type 1 or so type 3 above) the result is also type 3... but in general two cutouts of an extension create two extensions no matter where they happen.

For the second, if the input is type 2 (no data in primary) then metadata in the primary applies (inherits) to the extension and you have to create an MEF with the primary + one extension. If it is of type 3, you ignore do not output [0] and then the result does turn into a simple (so again: care with SIMPLE and NEXTEND if that happens in [0]).

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java
pdowler: mistakenly mistyped inherent instead of inherit in the earlier review (fixed now for clarity). Exentions in MEF files of type 2 inherit metadata from the primary (can't recall which has precedence if a header is set in both - I'd assume the most local but that could be wrong because of reasons).

at88mph:
I've added more rework as there were some edge cases failing the tests.  Please let me know if this is more what you were thinking with the comments above.

 - First pass through the file determines overlapping cutouts
 - Matched overlaps determine if output will be MEF or not
 - Types of MEFs laid out above should be handled
 - Some keywords were still missing.  Adjusted `XTENSION` and `SIMPLE` as needed.

pdowler:
I changed the CI build to build master of opencadc/nom-tam-fits and resolved a merge conflict in cadc-data-ops-fits/build.gradle in this PR using the web tool (which I think pushed a commit to your fork - it wasn't clear while doing it)... 

That didn't fix the build because the PR still depends on another nom-tam-fits API change you made to add `Fits.getHDU(name, version)`. I looked at your PR in the main nom-tam-fits repo but I don't think the correct navigation of the fits structure requires that approach and it has some poor behaviour when used with a stream (which we do have to be careful with when adding/changing Fits methods). 

Please try to rewrite without further nom-tyam-fits changes and I'd probably close that upstream PR since we'll make necessary changes in the opencadc fork for now. It should be possible to traverse the Fits object and find the required extensions (for now, by index or name/version and in future by which intersect a long/lat region), keep track of their indices, and output them (or parts of them) in the right order. For example, with input like 
```
pix=[2]&pix=[4]&pix=[3]
```
the output should be extensions 2, 3, 4 **in that order** because that's the order in the file. I am pretty sure this can be done without adding code to nom.tam.fits.Fits and I'm loathe to add convenience methods at this time.

at88mph:
I've removed the PR for NOM TAM and worked the logic into the `NDimensionalSlicer`.  I've been printing the number of HDUs read to confirm the minimal amount of reads are being read (i.e. for a cutout of `[0][100:300][0][400:475]` there is a single read).

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java
pdowler: **NDimensionalSlicer.getOverlap():**
This *must* iterator through the fits extensions in order and pick out the ones that overlap **in the order** they occur in the file. By using getHDU(fits, slice) you are jumping around in the fits file in the arbitrary request order. The order in the returned array will still end up being request order rather than file order. 

So the loop here must be calling `fits.readHDU()` until it finds all the requested extensions. It will have to read to the end if a requested slice is not in the file, which is a worst case. It is also important for this method to have a comment somewhere that states that the returned array has the extension slices in the correct -- file -- order.

Secondarily, it relies heavily on true random access with backwards seeking so could never be made to work (in a sub-optimal fashion) with a seekable stream. That's not an immediate use case but no sense in doing things that make that worse. Specifically, this code can be done in a way that it could work on a seekable stream with 2 passes through the stream. By allowing backwards seeks, that will be N passes instead of 2.. and they are not necessary and provide no benefit.

pdowler: **NDimensionalSlicer line 250:** wasn't the HDU already found by getOverlap()?

pdowler: **NDimensionalSlicer line 489 (hasData):**

Trying to trace what happens when this is called... `data` should be an ImageData and getSize() calls getTrueSize() which returns  the `byteSize` member. That gets assigned in one of the ctors and I assume in this usage it is the ImageData(Header) ctor, which means parseHeader() computes the byteSize from BITPIX, NAXIS, and NAXISn values. Can you verify that this is what happens?

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java
```diff
@@ -0,0 +1,587 @@
...
+        return overlapHDUIndexesSlices;
+    }
+
+    private Slices.ExtensionSliceValue[] getOverlap(final BasicHDU<?> hdu, final int hduIndex,
+                                                    final Slices.ExtensionSliceValue[] extensionSliceValues)
+            throws FitsException {
+        final List<Slices.ExtensionSliceValue> overlappingSlices = new ArrayList<>();
+
+        if (hdu != null) {
+            if (!(hdu instanceof ImageHDU)) {
```
pdowler, line 549: I wasn'ty going to review "correct output" because round trip tests, fitsverify, and eventually astronomers are better at that than code review can be, but since you moved this `if (ImageHDU)` in here I will comment. I agree that it has to be in here because errors have to be thrown and reported before opening output stream so that means in here.

This method **should not blindly** throw away no-image extensions. If the caller is only extracting an extension then all extension types can be output -- whether or not an extension can be returned depends on the applicability of the operation to the extension. That will be a combination of content and whatever code we implement. 

For now:
* extension extraction should work for all extensions, extension not found? **TBD but for now IllegalArgumentException**
* pixel cutouts should work for image extensions, **else: UnsupportedOperationException**

In near future:
* sky cutouts work for image extensions with complete-enough WCS, **else: skip**

In future:
* some sort of row,column (~pixel) cutouts could be implemented to work on FITS tables
* sky cutout could be implemented on FITS tables if columns contain coordinates (eg event lists)

pdowler (changes requested):

pdowler (approved):
This looks close to merge and test on a variety of data. I will refactor the public API (FitsOperation and classes it exposes) as it needs to be comprehensible and use terminology that will be obvious to potential users and collaborators (primarily VO community).

PR #43, opened 2021-01-19
refactoring to pull common param handling up to cadc-dali

pdowler:
and add prototype cfitsio-style extension and pixel cutout handling to SODA impl


pdowler:
fixing CI build issues...

cadc-dali/build.gradle
```diff
@@ -15,7 +15,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '1.2.3'
+version = '1.2.4'
```
at88mph, line 18: It's true there are no backward incompatibilities, but the new features could push this to 1.3.

at88mph (changes requested):

cadc-dali/build.gradle
pdowler: Yes, that is what semantic versioning (semver.org) says... but that doesn't seem a good match for how we work.

We tend to make many small increments where we add a class or method here and there. Plus, in working with ESAC on caom2 code we agreed to try constraining versions more tightly (they requested it) in the build files, so in core and caom2 and caom2db repos you will see pretty tight dependencies like `[2.4,2.5)`... this is to prevent ever running into an incompat change. I'm not saying that was a good idea itself, but it does imply a certain meaning to versioning that isn't the same as semver. It also makes me not want to increment the minor version for source- and binary-compatible changes.

The policy in caom2 is more like:
* major version: lots of work to adapt due to conceptual/design changes
* minor version: local compile incompatibility, easy to fix
* patch version: compile- and binary-compatible (even with a ton of new "features")

Is that the best? Probably not, but neither is version 1.87.3

Here, if new feature is 1.3 then even the smallest incompat change is 2.0; that doesn't fit how we work. The one you didn't catch is cadc-soda-server, which I only incremented 1.1.2 to 1.1.3 instead of 1.2.0 :-)

at88mph (approved):

PR #46, opened 2021-02-09
Added FITS Header WCS Keyword collector.

at88mph:


cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/FITSHeaderWCSKeywords.java
```diff
@@ -0,0 +1,161 @@
...
+import nom.tam.fits.Header;
+import nom.tam.fits.HeaderCard;
+import nom.tam.fits.header.Standard;
+import org.apache.log4j.Logger;
+
+
+/**
+ * Create WCS Keywords class from a FITS header.  Conversions are limited to the underlying data types.
+ */
+public class FITSHeaderWCSKeywords extends WCSKeywordsImpl {
```
pdowler, line 89: It would be better to implement the interface rather than use this map-based impl. There isn't much point in copying all the keywords to a map when the interface methods map to `Header` methods ore or less directly... that's actually the intent of the interface.

I think from the comment that the map-based impl is something poorly extracted from caom2.

pdowler (changes requested):

at88mph:
Rework submitted.

pdowler (changes requested):
Since this WCSKeywords wrapper is to support computing metadata, it wraps the input fits header. I think you could make all the `put` methods throw UnsupportedOperationException to help prevent unintended use.... worry: someone uses it with a local file that is writable, and someone else "optimises" something in cadc-wcs and suddenly file gets modified (see below).

IIRC `put` is part of the API to support cadc-wcs `Transform.translate()` and that makes a copy and returns a map-based WCSKeywordsImpl so it doesn't modify the input keywords (so didn't really need to be in the interface and WCSKeywordsImpl didn't need to be public, but I think that's a result of being pulled put of caom2). That does mean the iterator needs to work :-)

at88mph:
That's a good idea and is done.

pdowler (approved):

PR #49, opened 2021-03-17
Base cutout classes

at88mph:
This is a smaller PR to have some setup in place for the concrete SODA cutouts to take place.
 - FITSHeaderWCSKeywords class was reworked to allow copying the input Header to ensure nothing modifies the original one.  Also adds a sanitization step for it, as well as Trace observability.
 - Add base cutout classes and tools to enable coming concrete Shape/Energy cutouts.


.github/workflows/gradle.yml
```diff
@@ -30,7 +31,9 @@ jobs:
       run: cd cadc-datalink-server && ../gradlew --info clean build javadoc checkstyleMain install
       
     - name: build and test cadc-soda-server
-      run: cd cadc-soda-server && ../gradlew --info clean build javadoc checkstyleMain install
```
pdowler, line 33: Why is cadc-soda-server build removed?

.github/workflows/gradle.yml
```diff
@@ -30,7 +31,9 @@ jobs:
...
-      
+
+    - name: install WCSLIB (libwcs7)
```
pdowler, line 35: The published cadc-wcs only supports wcslib5; I came up with a scheme to include JNI binding lib for multiple wcslib versions, but the grunt work to build and test for wcslib6 and 7 hasn't been done. Will create an issue for that work.

cadc-data-ops-fits/build.gradle
```diff
@@ -8,19 +8,27 @@ plugins {
 repositories {
     mavenCentral()
     mavenLocal()
+
+    // Remove when cadc-soda-server:1.2.1 is on mavenCentral()
```
pdowler, line 12: What is this? That version is on mavenCentral and since the first mavenCentral version was 1.2.0 this should never have been published to jcenter.

at88mph:
Comments addressed.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/CoordTypeCode.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+package org.opencadc.fits.slice;
+
+import java.util.Arrays;
+import java.util.Locale;
+
+
+public enum CoordTypeCode {
+    // Spatial type codes.
```
pdowler, line 76: I would not include the hyphen spacers in the literal values here; the startsWith in getdefaultUnit should still match. It's feasible for any ctype to be linear and thus not have a projection  algorithm, in which case the stype might just be "RA" (at least in valid test code, maybe not in practice).

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/CoordTypeCode.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    public static String getDefaultUnit(final String ctype) {
+        if (ctype == null) {
+            return null;
+        }
+
+        final CoordTypeCode matchedCoordTypeCode =
+                Arrays.stream(values()).filter(coordTypeCode -> ctype.toUpperCase(Locale.ROOT).substring(0, 4).startsWith(
+                        coordTypeCode.typeCodeString)).findFirst().orElse(null);
```
pdowler, line 122: Since this uses startsWith, you don't need the substring(0,4) here and in principle it could fail if the ctype is short which is technically valid (as noted in previous comment).

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/FITSCutout.java
```diff
@@ -0,0 +1,102 @@
...
+    protected FITSCutout(final FITSHeaderWCSKeywords fitsHeaderWCSKeywords) {
+        DaliUtil.assertNotNull("fitsHeaderWCSKeywords", fitsHeaderWCSKeywords);
+        this.fitsHeaderWCSKeywords = fitsHeaderWCSKeywords;
+    }
+
+
+    /**
+     * Obtain the bounds of the given cutout.
+     * @param cutoutBound   The bounds (shape, interval etc.) of the cutout.
+     * @return  long[] array of overlapping bounds, or long[0] if all pixels are included.
```
pdowler, line 94: Does this return null if no pixels are included? The caom2 cutout code had those 3 return options... or something else?

cadc-data-ops-fits/build.gradle
```diff
@@ -14,13 +14,18 @@ sourceCompatibility = 1.8
 group = 'org.opencadc'
 version = '0.1.2'
 
+// To allow for french extended ASCII characters in file javadoc.
+javadoc {
```
pdowler, line 18: We've never seen a need for this before... are the offending characters from the french in the copyright header? Is this another thing caused by gradle 7 or does it also effect CI with gradlew and gradle 6.8.3? 

We have so many small modules that I don't like to solve problems with something in every build file if I can possibly avoid it... please remove it here and create an issue instead.

.github/workflows/gradle.yml
```diff
@@ -1,6 +1,7 @@
 name: Java CI
 
-on: 
+on:
+  push:
```
pdowler, line 4: on push is not needed; CI intentionally triggered for pull request actions only

pdowler (changes requested):

.github/workflows/gradle.yml
at88mph: OK.  I like that it runs on my fork before going as it finds the occasional thing (i.e. `checkstyleMain` errors) that may have been missed.  Removed.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/FITSCutout.java
at88mph: Returns null, like CAOM-2 does.

at88mph:
Comments addressed.

pdowler (approved):

PR #53, opened 2021-04-14
Fix Gradle build files for Gradle 7 (CI)

at88mph:


pdowler:
using gradlew and 6.8.3

PR #55, opened 2021-05-18
Spatial cutouts

at88mph:
The vast majority of the files are header text files for testing.  Using these alleviates the need for including (or acquiring) FITS files for the tests.


pdowler (changes requested):
The PolygonCutout class seems to be missing the `PolygonCutout(FITSHeaderWCSKeywords)` constructor that would enable CircleCutout and RangeCutout to not have to unwrap (and PolygonCutout's call to super to re-wrap) the Header. The unwrap/rewrap is where those create a `new PolygonCutout`.

I don't see the BaseCutoutTest... wanted to look at assertFuzzyPixelArrayEquals. Probably not added... and I'd like to see the CI bukld complete (which will need that)

at88mph:
The `BaseCutoutTest` was added in the last PR (`base-cutout`) in preparation for the tests that would use it.  That file has already been merged into https://github.com/opencadc/dal/blob/master/cadc-data-ops-fits/src/test/java/org/opencadc/fits/slice/BaseCutoutTest.java.

at88mph:
I've updated the `PolygonCutout` class.  The CI build passes now.  It failed three weeks ago due to some GitHub Actions internal error.

pdowler (approved):

PR #56, opened 2021-06-09
Spectral cutouts

at88mph:
Added EnergyCutout
Added HDUIterator for ease of walking HDUs
Added PixelCutout to maintain XXXCutout API


cadc-data-ops-fits/src/main/java/org/opencadc/fits/HDUIterator.java
```diff
@@ -0,0 +1,172 @@
...
+
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import nom.tam.fits.BasicHDU;
+import nom.tam.fits.Fits;
+import nom.tam.fits.FitsException;
+
+
```
pdowler, line 79: Surprised to see this class show up only now since you already implemented spatial cutouts without it.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyConverter.java
```diff
@@ -0,0 +1,156 @@
...
+ ************************************************************************
+ */
+
+package org.opencadc.fits.slice;
+
+import ca.nrc.cadc.util.ArrayUtil;
+
+import org.apache.log4j.Logger;
+
+
```
pdowler, line 75: EnergyConverter looks like a fork of the class in caom2, but lacking energy support (eV and prefixes)...

It would be better to move the class from caom2 to someplace upstream and augment it with new methods and units and multiplicative factors as necessary.

**move**: copy it from caom2 to cadc-dali (package ca.nrc.cadc.dali) and augment; I'll make a note to remove it from caom2... there are other things in caom2 that are gradually migrating to cadc-dali so I'll add this to the caom2 cleanup list

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/WCSCutoutUtil.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+
+import nom.tam.fits.Header;
+import nom.tam.fits.HeaderCardException;
+import org.opencadc.soda.server.Cutout;
+
+
+public class WCSCutoutUtil {
+    public static long[] getBounds(final Header header, final Cutout cutout)
+            throws HeaderCardException, NoSuchKeywordException {
```
pdowler, line 87: This only does cutout of one type: the first one found in hard coded order. In SODA, the caller can specify cutout along multiple axes and they should be combined, eg
```
CIRCLE=12 34 0.1 & BAND=200e-9 400e-9 & TIME=54321.5 54321.6 & POL=I & POL=Q
```
(extra space for clarity) in a 5-D cube. You can treat the axes as orthogonal so what I saw of the spatial cutouts didn't obviously make me think this was what you had in mind... the caom2 cutout code does each axis separately and recombines the long[] pixel ranges into the single n-D pixel cutout action so the FITSCutout and subclass(es) approach can work, but not as simply as this.

pdowler (changes requested):
It would be good for each class to have a class-level javadoc with a brief statement of what the class does. If I didn't mention that before, it's probably because the class(es) with missing comment were obvious enough and I only notice it when it isn't so obvious. 

More detailed comments below...

at88mph:
Rework complete.  The use case of multiple WCS cutouts is supported for 0-1 cutouts per axis.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/WCSCutoutUtil.java
pdowler: I had to dig very deep to see how this is intended to work - I had missed the change in FITSCutout.getBounds behaviour where it returns a long[NAXIS] rather than a long[n] or long[0]. The javadocs don't consistently describe the contract of that method. 

Specifically in PolygonCutout, lines ~238-248: there is an assumption in the array copy and fill that the spatial axes are 1 and 2 but that doesn't have to be the case. Energy could be 1 and spatial could be 2 and 3. So that needs to be generalised and WCSCutoutUtilTest needs tests with axes in different orders.

EnergyCutout lines~161-175 do not make that assumption and are a better model of how to put the clipped bounds into the right place in the entireBounds array.

pdowler: It's a bit of a shame that the clipped bounds -> entire bounds happens in multiple Cutout<T> implementations; as such, each should have a test where the axis in question is in an atypical place in the order...

PolygonCutoutTest probably needs to be updated with a test where spatial axes are not 1,2... and if you don't have it, I'd also have a test where DEC axis comes before RA (we do have some data like that in the archive, hence the CoordSys.swappedAxes) because you might need to take that into account when doing clipped -> entire bounds.

The EnergyCutoutTest(s) all seem to have energyAxis == 3 (which is typical for cube). I would at least add a test for 1-D spectra so that energyAxis == 1 (also typical).

pdowler (changes requested):

cadc-data-ops-fits/src/main/java/org/opencadc/fits/CADCExt.java
```diff
@@ -80,16 +80,19 @@
...
     PC2_1(HDU.IMAGE, VALUE.REAL, ""),
     PC2_2(HDU.IMAGE, VALUE.REAL, ""),
     RESTFRQ(HDU.IMAGE, VALUE.REAL, ""),
 
     /**
      * Use RESTFRQ
+     * @deprecated Use RESTFRQ
      */
-    @Deprecated
+    @Deprecated()
```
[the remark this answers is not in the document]
at88mph: Ha!  No.  I thought one could write a deprecated message as an argument but I never removed the parentheses when I realized that I couldn't do that.

at88mph:
All review comments should be accounted for now.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/PolygonCutout.java
```diff
@@ -189,23 +193,19 @@ public PolygonCutout(FITSHeaderWCSKeywords fitsHeaderWCSKeywords) {
...
-
-            for (int i = spatialCoords.length; i < worldCoords.length; i++) {
+            final double[] worldCoords = new double[naxis];
+            // Set the spatial values where they should be.
+            worldCoords[spatialLongitudeAxisIndex] = point.getLongitude();
+            worldCoords[spatialLatitudeAxisIndex] = point.getLatitude();
+
+            // Fill in the rest of the world coordinates.
+            for (int i = 0; i < worldCoords.length
+                            && i != spatialLongitudeAxisIndex && i != spatialLatitudeAxisIndex; i++) {
```
pdowler, line 208: Are you sure this loop does what you want?
```
groovy:000> for (int i=0; i<5 && i != 3; i++) { System.out.println(i); }
0
1
2
===> null
```
says it would only fill in `worldCoords` up to but not including `spatialLongitudeAxisIndex` or `spatialLatitudeAxisIndex` but not after.

pdowler (changes requested):

at88mph:
Yes, good catch.  It would've broken out of the loop at the incorrect time as well.

pdowler (approved):

PR #58, opened 2021-07-08
Polarization cutouts

at88mph:


cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/PolarizationState.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ ************************************************************************
+ */
+
+package org.opencadc.fits.slice;
+
+/**
+ * Represents the various Polarization state values to be used when calculating cutouts.
+ */
+public enum PolarizationState {
```
pdowler, line 74: This is from caom2 but a good improvement from the 2011 code to use the enum mechanics to encode the names-number mapping :-)

For general use, this class should be up in cadc-dali (main dali pkg is fine). And it needs to have a static methods for both int -> PolarizationState and String -> PolarizationState. The latter will be needed when I remove the class from caom2 and it can be used in CommonParamValidator.validatePOL to validate input. Early validation of input is always a win.

That validatePOL method signature could be changed to return List<PolarizationState>, which would in turn change the type in the soda Cutout class... that will continue to trickle down into other code so it's a bit of a pain. But String all through a set of APIs usually leads to obscure bugs. Maybe leave that change for last and  see how painful it might be... problem with String is that it gets used in ways that the compiler won't detect (because of automagic toString() but also equals).

cadc-data-ops-fits/src/test/java/org/opencadc/fits/slice/PolarizationCutoutTest.java
```diff
@@ -0,0 +1,226 @@
...
+        final long[] results = testSubject.getBounds(states);
+        final long[] expected = new long[]{1L, 300L, 1L, 300L, 1L, 151L, 1L, 1L};
+
+        assertFuzzyPixelArrayEquals("Wrong output.", expected, results);
+    }
+
+    @Test
+    @Ignore("Would this ever happen?")
+    public void testNoOverlap() throws Exception {
+
```
pdowler, line 224: Yes, you can have a cube with IQU on the polarization axis and the caller asks for RR. Simple test that should be included.

pdowler (changes requested):

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/PolarizationState.java
at88mph: For a String -> PolarizationState, is this the same as `Enum.valueOf(name)`?

pdowler: Could be, some enums in caom2 have values that are not valid enum names ("X-ray") hence the explicit method that loops over values():
```
    public static PolarizationState toValue(String s) {
        for (PolarizationState ps : values()) {
            if (ps.value.equals(s)) {
                return ps;
            }
        }
        throw new IllegalArgumentException("invalid value: " + s);
    }
```
which also gives you control over the exception and message. You would have to change the check to use the name since value was the String in the above snippet.

at88mph:
All rework submitted.  This PR now spans `cadc-dali`, `cadc-sia`, `cadc-soda-server`, and `cadc-data-ops-fits`.

pdowler (approved):

PR #67, opened 2021-07-27
Modified FitsOperations to use Cutout class.  Added test for HDU Iter…

at88mph:
…ator to increase code coverage.

### DIFF
--- a/cadc-data-ops-fits/build.gradle
+++ b/cadc-data-ops-fits/build.gradle
@@ -24,7 +24,7 @@ dependencies {
     compile 'org.opencadc:cadc-soda-server:[1.2.1,2.0)'
     compile 'org.opencadc:cadc-wcs:[2.0,3.0)'
     compile 'org.opencadc:jsky:[1.0.0,2.0.0)'
-    compile 'org.opencadc:nom-tam-fits:[1.16.1,1.17.0)'
+    compile 'org.opencadc:nom-tam-fits:[1.16.2,1.17.0)'
 
     // Use JUnit test framework
     testCompile 'junit:junit:[4.13,5.0)'
--- a/cadc-data-ops-fits/build.gradle
+++ b/cadc-data-ops-fits/build.gradle
@@ -24,7 +24,7 @@ dependencies {
     compile 'org.opencadc:cadc-soda-server:[1.2.1,2.0)'
     compile 'org.opencadc:cadc-wcs:[2.0,3.0)'
     compile 'org.opencadc:jsky:[1.0.0,2.0.0)'
-    compile 'org.opencadc:nom-tam-fits:[1.16.2,1.17.0)'
+    compile 'org.opencadc:nom-tam-fits:[1.16.2,1.16.3)'
 
     // Use JUnit test framework
     testCompile 'junit:junit:[4.13,5.0)'
--- a/cadc-data-ops-fits/build.gradle
+++ b/cadc-data-ops-fits/build.gradle
@@ -12,7 +12,7 @@ repositories {
 
 sourceCompatibility = 1.8
 group = 'org.opencadc'
-version = '0.2.2'
+version = '0.2.3'
 
 description = 'OpenCADC FITS cutout library'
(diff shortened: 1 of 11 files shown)


cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NoOverlapException.java
```diff
@@ -0,0 +1,112 @@
...
+*
+************************************************************************
+*/
+
+package org.opencadc.fits.slice;
+
+/**
+ * Exception thrown to indicate a client error in the supplied cutout.
+ */
+public class NoOverlapException extends IllegalArgumentException {
```
pdowler, line 74: If I saw all the uses, this is thrown by NDimensionalSlicer but for the purposes of FitsOperations it is just an IllegalArgumentException? Just wondering if FitsOperations should declare the throws (and document) and thus if this exception should be in the parent package with FitsOperations... it won't matter for our use in minoc and alma because we currently treat it like an IllegalArgumentException (400) but I could see an argument to be made for knowing to catch and handle this differently. It's not like a "resource not found" (404) but someone could argue that it's more like not finding a matching item in a collection... like maybe we would want to change to 204 "No Content" to say "what you asked for was valid, just didn't find any matching pixels". 

If we put this exception in the parent package and declare the throws from FitsOperations, then we could change it to not be an IllegalArgumentException much more easily. I think it's a prudent change.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/FitsOperations.java
```diff
@@ -156,20 +159,21 @@ public void headersToStream(final OutputStream outputStream) throws IOException
...
-     * @param slices subsets of data to extract
+     * @param cutout cutout spec
      * @param outputStream  The Stream to write out to.
-     * @throws ReadException    Any errors to report back to the caller.
+     * @throws ReadException        Any errors reading reported by the storage system.
+     * @throws NoOverlapException   Invalid cutout as there is no match (overlap).
      */
-    public void cutoutToStream(final List<ExtensionSlice> slices, final OutputStream outputStream) throws ReadException {
+    public void cutoutToStream(final Cutout cutout, final OutputStream outputStream) throws ReadException {
+        log.debug("cutoutToStream() start.");
```
pdowler, line 168: There should be a check for not-null args and maybe valid Cutout here at the entry point. (just because this is a general purpose library someone could use outside alma and minoc)

See https://github.com/opencadc/dal/issues/68 for a way to make validity checking a one-liner.

pdowler (changes requested):

cadc-data-ops-fits/src/main/java/org/opencadc/fits/FitsOperations.java
at88mph: The requirements in #68 could be met by what the new `SodaCutout` inner class will look like in the next PR for Minoc's `GetAction`:
https://github.com/at88mph/storage-inventory/blob/cadc_8786/minoc/src/main/java/org/opencadc/minoc/GetAction.java

Adding a `validate()` method can simple ensure Its `getConflicts()` method will return an empty `List`.  In fact, much of that `SodaCutout` class could go into this `Cutout` class.

cadc-data-ops-fits/build.gradle
```diff
@@ -24,7 +24,7 @@ dependencies {
     compile 'org.opencadc:cadc-soda-server:[1.2.1,2.0)'
     compile 'org.opencadc:cadc-wcs:[2.0,3.0)'
     compile 'org.opencadc:jsky:[1.0.0,2.0.0)'
-    compile 'org.opencadc:nom-tam-fits:[1.16.1,1.17.0)'
+    compile 'org.opencadc:nom-tam-fits:[1.16.2,1.16.3)'
```
pdowler, line 27: I think you wanted to change the lower bound here

pdowler (approved):

PR #72, opened 2021-11-26
CADC-10517: first checkin of cadc-pkg-server

hjeeves:


cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
```diff
@@ -0,0 +1,82 @@
...
+import java.net.URL;
+
+/**
+ * Class to describe a file to be added to a Zip or Tar package.
+ */
+public class PackageItem
+{
+    public URL url;
+    public String filename;
+}
```
pdowler, line 82: With the current scope, these fields cannot be null, so they should be private and final (public indicates "may be null" to other code). I would use `relativePath` instead of `filename` as a reminder of what's allowed.

There should be a constructor `PackageItem(String relativePath, URL url)` that enforces not-null

Then you would need `getRelativePath()` and `getURL()` to provide access.

Complete javadocs for ctor. Javadoc for get methods: not really necessary but fine if included.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,260 @@
...
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.net.URI;
+import java.net.URL;
+import java.util.Date;
+import java.util.Iterator;
+import org.apache.log4j.Logger;
+
+public abstract class PackageRunner implements JobRunner
```
pdowler, line 95: This now needs a class javadoc to explain what it is.

Opening brace && code style: most packages in dal.git have javadoc and checkstyleMain targets invoked in the CI build. I expect both to be in place for this module and results to be silent (not even warnings from javadoc for "new" code... this is kind of new code)

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,260 @@
...
+
+    public PackageRunner() 
+    {
+    }
+
+    /**
+     * Generate the package name
+     * @return String with package name.
+     */
+    protected abstract String generatePackageName();
```
pdowler, line 114: `getPackageName()`??

Also (relateed to this method): you are going to want code like that in `initTarResponse()` to set headers, inc. calling getPackageName() for the `content-disposition` header. I mention it here because PackageRunner being responsible for setting headers before opening stream is the reason this abstract method exists.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java
```diff
@@ -0,0 +1,257 @@
...
+        List<TarContent> t = map.get(path);
+        if (t == null)
+        {
+            t = new ArrayList<TarContent>();
+            map.put(path, t);
+        }
+        t.add(tc);
+    }
+    
+    public void write(String path, URL url)
```
pdowler, line 173: Should change this method signature to `write(PackageItem)` to show how it's connected to the other classes. It's true that this class would be slightly more generic and still usable without doing that -- the caller could just do
```
write(item.getRelativePath(), item.getURL());
```
However, we found out in cadc-soda-server that when we did something like that we couldn't add extra info without changing the arg list of the method, so it was easier to evolve if we passed a struct and then had the struct evolve.

cadc-pkg-server/src/test/java/org/opencadc/pkg/server/PackageRunnerTest.java
```diff
@@ -0,0 +1,116 @@
...
+public class PackageRunnerTest {
+    private static final Logger log = Logger.getLogger(PackageRunnerTest.class);
+
+    static {
+        Log4jInit.setLevel("ca.nrc.cadc.caom2", Level.INFO);
+    }
+    
+    public PackageRunnerTest() { 
+    }
+
```
pdowler, line 90: Delete the commented out lines, but let's keep the test class for now because we might want to implement some test code here later.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,260 @@
...
+                logInfo.setSuccess(false);
+                logInfo.setMessage("Could not set job phase to executing, was: " + ep);
+                return;
+            }
+            log.debug(job.getID() + ": QUEUED -> EXECUTING [OK]");
+
+            // TODO: there will need to be something here to write the
+            // Items out to a tar file.
+            Iterator<PackageItem> packageItems = getItems();
+
```
pdowler, line 182: In principle, once you address the other comments it is possible to write the output loop right here, something like:
```
while (packageItems.hasNext()) {
    PackageItem pi = packageItems.next();
    writer.write(pi);
}
```
This would show/verify that the api elements all fit together.

pdowler (changes requested):
Please add this module to .github/workflows/gradle.yml as well.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: - 'generatePackageName' is from the original code, as the filename is 'generated' from using a static string + the job id. (cadc-download-<jobid>.tar) - Doesn't matter if it's get instead...

- content-disposition header content will depend on what kind of package is being written.  
tar: Content-Disposition = "inline;filename=*.tar"
zip: Content-Disposition="attachment;filename=*.zip"

... if I've read about this correctly. At the least the file extension will differ. 

So either the whole function is provided by extending classes, or there's variables that need to be populated. Which way to go?

pdowler: In last week's slack thread we said `getPackageName()`. 
I would say subclass (impl) should provide the base filename (sans extension) and it can decide to include something to avoid filename collisions (jobid, timestamp, or something  meaningful like caom2-packlage-server did when there was a single input ID). 
PackageRunner should append the correct extension consistent with the `content-type`.

hjeeves: The Apache Commons Compress library automatically generates an extension for any files it generates. 

Content disposition filename extension needs to match that. So it's a value that should come from the Writer class (TarWriter or ZipWriter.)

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: I can add this as code with a comments, but as the function sits it would never work - there's no place that the TarWriter is instantiated yet.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
hjeeves: done

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: added checkstyleMain to build.gradle, and a javadoc for the class. - gradle javadoc and gradle checkstyleMain both pass now

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java
hjeeves: done

cadc-pkg-server/src/test/java/org/opencadc/pkg/server/PackageRunnerTest.java
hjeeves: done

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: I believe there's enough function calls and documentation to show the Steps above in the PackageRunner doIt() function now.  Added functions to instantiate the syncOutput stream so TarWriter can be used. Remains to be seen how it acts when really used.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,291 @@
...
+                return;
+            }
+            log.debug(job.getID() + ": QUEUED -> EXECUTING [OK]");
+
+            // Build an iterator of PackageItem from the files named
+            // in the local Job instance
+            Iterator<PackageItem> packageItems = getItems();
+
+            // TarWriter needs an output stream instantiated.
+            initOutputStream(getPackageName());
```
pdowler, line 181: the `new TarWriter` needs to be assigned to a variable outside the try/catch in order for the finally to be able to close it.

In the original code it was not a class member variahble,  just a local var declared before the try. It's always best practice to declare as locally as possible, so up on line~163

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,291 @@
...
+
+    /**
+     * Initialize syncOutput output stream with the correct content type and disposition,
+     * as provided by the writer class
+     * @param packageName - package name to use
+     * @throws IOException
+     */
+    private void initOutputStream(String packageName) throws IOException {
+        syncOutput.setResponseCode(200);
+
```
pdowler, line 245: Either here or up where getPackageName() is called, there needs to be a check for null. I'd normally put it where getPackageName() is called and to keep run() clear I would probably make this method no-arg and call getPackageName() In here.

So, if the subclass returns null, what to do - fail or default value? 
Fail:
```
throw new RuntimeException("BUG: getPackageName() returned null");
```
or default value:
```
packageName = "default-" + job.getID();
```

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+    
+    private class TarContent {
+        URL url;
+        String filename;
+        String contentMD5;
+        String emsg;
+    }
+    
+    public void close() throws IOException {
```
pdowler, line 113: most of this should be removed; just keep the last 2 lines (flush and close)

pdowler (changes requested):

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java
hjeeves: ok - removed everything that writes the README file out.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: Opted to put a check in initOutputStream, at the least. Left PackageRunner with no default package name.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: done

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,291 @@
...
+    protected JobUpdater jobUpdater;
+    protected SyncOutput syncOutput;
+    protected ByteCountOutputStream outputStreamCounter;
+    protected WebServiceLogInfo logInfo;
+    protected TarWriter writer;
+
+    public PackageRunner() {}
+
+    /**
+     * Get filename for package to be created.
```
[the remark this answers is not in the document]
hjeeves: changed text.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,291 @@
...
+ * To make a child class, implement these:
+ * - Iterator PackageItem getItems(): Will be used to download files to the package
+ * - String getPackageName(): filename used to create the package.
+ * </p>
+ */
+public abstract class PackageRunner implements JobRunner {
+    private static final Logger log = Logger.getLogger(PackageRunner.class);
+
+    protected Job job;
+
```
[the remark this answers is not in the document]
hjeeves: done

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -0,0 +1,291 @@
...
+import org.apache.log4j.Logger;
+
+/**
+ * Abstract class PackageRunner: a JobRunner implementation that provides core
+ * functionality for building CADC package files (zip or tar files, for example.)
+ * Job state management and error reporting back to the Job are provided.
+ * <p>
+ * To make a child class, implement these:
+ * - Iterator PackageItem getItems(): Will be used to download files to the package
+ * - String getPackageName(): filename used to create the package.
```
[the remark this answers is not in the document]
hjeeves: removed second paragraph of text.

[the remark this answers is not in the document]
hjeeves: made edits to PackageItem as requested.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: changed to getPackageName(), added a function to set up the OutputStream with the correct disposition string, content-type is hard coded in PackageRunner for now as only tar is currently supported. It's expected that will change when a mechanism to support requesting tar or zip is implemented.

pdowler (approved):

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java
```diff
@@ -122,13 +122,18 @@ public void write(PackageItem packageItem) throws TarProxyException, IOException
                 log.info("HEAD " + packageURL.toExternalForm() + " failed, reason: " + get.getThrowable().getMessage());
                 throw new TarProxyException("HEAD " + packageURL.toExternalForm() + " failed", get.getThrowable());
             }
-            String filename =  packageItem.getRelativePath();
+
+            // Get the filename that comes back wiht the download
+            String filename = get.getFilename();
+            // Prepend the relative path for the item so the tar structure is
+            // built correctly
+            String archiveFilename =  packageItem.getRelativePath() + "/" + filename;
```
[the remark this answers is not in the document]
hjeeves: In the original TarWriter, lines 193-194, the filename for an artifact is pulled from the HEAD request response. It's not available prior.  <snip>

After discussions, it's decided that the PackageItem relativePath will be the scheme specific part of an individual Artifact uri. The portion of the file path that was derived from the Plane URI will not be used inside the tar files.

PR #76, opened 2022-01-29
Adapt to latest vocabulary.

### LINKED ISSUE #75: Use latest vocabulary
The latest vocabulary was updated on January 27th, 2022.

https://www.ivoa.net/rdf/datalink/core

Part of the drive here would be to make the OpenCADC vocabulary items (e.g. `thumbnail` and `package`) part of the core vocabulary for other institutions.

at88mph:
Fixes #75

### DIFF
--- a/cadc-datalink-server/src/main/java/org/opencadc/datalink/DataLink.java
+++ b/cadc-datalink-server/src/main/java/org/opencadc/datalink/DataLink.java
@@ -88,13 +88,6 @@ public class DataLink {
      */
     public enum Term { // TODO: re-use the VocabularyTerm code once extracted from caom2
         THIS("#this"),
-
-        /**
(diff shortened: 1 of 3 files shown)


cadc-datalink-server/src/main/java/org/opencadc/datalink/DataLink.java
```diff
@@ -89,13 +89,16 @@
     public enum Term { // TODO: re-use the VocabularyTerm code once extracted from caom2
         THIS("#this"),
 
+        /**
+         * @deprecated See https://www.ivoa.net/documents/DataLink/20150617/REC-DataLink-1.0-20150617.html#toc9 for
+         *      properly adding a recursive DataLink entry.
+         */
```
pdowler, line 95: The #datalink should probably just be removed since the idea was rejected.

pdowler (changes requested):

pdowler (approved):

PR #77, opened 2022-02-02
ZipWriter support, RESPONSEFORMAT job param to select pkg type: CADC-10473

hjeeves:
Updates to work adamian did for initial ZipWriter. Added check for RESPONSEFORMAT in the incoming job to separate this out from updating caom2-pkg-server to have a RESPONSEFORMAT parameter.

### DIFF
--- a/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
+++ b/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
@@ -3,7 +3,7 @@
  *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
  **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
  *
- *  (c) 2022.                            (c) 2022.
+ *  (c) 2021.                            (c) 2021.
  *  Government of Canada                 Gouvernement du Canada
  *  National Research Council            Conseil national de recherches
  *  Ottawa, Canada, K1A 0R6              Ottawa, Canada, K1A 0R6
--- a/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
+++ b/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
@@ -83,17 +83,14 @@ import java.net.URL;
 public class PackageItem {
     private final URL url;
     private final String relativePath;
-    private final boolean isDirectory;
 
     /**
      * Instantiate a PackageItem object.
      * @param url - URL where a file can be accessed for download.
      * @param relativePath - Relative path of the file referenced by url parameter.
      *                     Used to build the correct directory structure in the final package.
-     * @param isDirectory - whether the item represents a directory
      */
-    public PackageItem(URL url, String relativePath, boolean isDirectory) {
-
+    public PackageItem(URL url, String relativePath) {
         if (url == null) {
             throw new IllegalArgumentException("parameter url required.");
         }
@@ -101,14 +98,11 @@ public class PackageItem {
         if (!StringUtil.hasText(relativePath)) {
             throw new IllegalArgumentException("parameter relativePath required.");
         }
-        this.isDirectory = isDirectory;
+
         this.url = url;
         this.relativePath = relativePath;
     }
 
-    public PackageItem(URL url, String relativePath) {
-        this(url, relativePath, false);
-    }
 
     public URL getURL() {
         return url;
@@ -117,9 +111,4 @@ public class PackageItem {
     public String getRelativePath() {
         return relativePath;
     }
-
-    public boolean isDirectory() {
-        return isDirectory;
-    }
-
 }
--- a/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
+++ b/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
@@ -106,7 +106,9 @@ public class PackageItem {
         this.relativePath = relativePath;
     }
 
-    public PackageItem(URL url, String relativePath) { this(url, relativePath, false); }
+    public PackageItem(URL url, String relativePath) {
+        this(url, relativePath, false);
+    }
 
     public URL getURL() {
         return url;
--- a/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
+++ b/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
@@ -3,7 +3,7 @@
  *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
  **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
  *
- *  (c) 2021.                            (c) 2021.
+ *  (c) 2022.                            (c) 2022.
  *  Government of Canada                 Gouvernement du Canada
  *  National Research Council            Conseil national de recherches
  *  Ottawa, Canada, K1A 0R6              Ottawa, Canada, K1A 0R6
@@ -83,14 +83,16 @@ import java.net.URL;
 public class PackageItem {
     private final URL url;
     private final String relativePath;
+    private final boolean isDirectory;
 
     /**
      * Instantiate a PackageItem object.
      * @param url - URL where a file can be accessed for download.
      * @param relativePath - Relative path of the file referenced by url parameter.
      *                     Used to build the correct directory structure in the final package.
+     * @param isDirectory - whether the item represents a directory
      */
-    public PackageItem(URL url, String relativePath) {
+    public PackageItem(URL url, String relativePath, boolean isDirectory) {
 
         if (url == null) {
             throw new IllegalArgumentException("parameter url required.");
@@ -99,10 +101,12 @@ public class PackageItem {
         if (!StringUtil.hasText(relativePath)) {
             throw new IllegalArgumentException("parameter relativePath required.");
         }
-
+        this.isDirectory = isDirectory;
         this.url = url;
         this.relativePath = relativePath;
     }
+
+    public PackageItem(URL url, String relativePath) { this(url, relativePath, false); }
 
     public URL getURL() {
         return url;
@@ -110,6 +114,10 @@ public class PackageItem {
 
     public String getRelativePath() {
         return relativePath;
+    }
+
+    public boolean isDirectory() {
+        return isDirectory;
     }
 
 }
--- a/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
+++ b/cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
@@ -91,12 +91,11 @@ import org.apache.log4j.Logger;
 /**
  * Abstract class PackageRunner: a JobRunner implementation that provides core
  * functionality for building CADC package files (zip or tar files, for example.)
+ * To select a package type, RESPONSEFORMAT must be provided as a job parameter.
  * Job state management and error reporting back to the Job are provided.
  */
 public abstract class PackageRunner implements JobRunner {
     private static final Logger log = Logger.getLogger(PackageRunner.class);
-    public static final String TAR_TYPE = "application/x-tar";
-    public static final String ZIP_TYPE = "application/zip";
 
     private JobUpdater jobUpdater;
     private SyncOutput syncOutput;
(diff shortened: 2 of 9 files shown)


cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -241,31 +233,49 @@ private void doIt() {
...
+            throw new IllegalArgumentException("packageName not provided");
         }
 
-        syncOutput.setResponseCode(200);
+        // Package type is in RESPONSEFORMAT Job parameter
+        String responseFormat = ParameterUtil.findParameterValue("RESPONSEFORMAT", job.getParameterList());
+
+        if (!StringUtil.hasLength(responseFormat)) {
+            throw new IllegalArgumentException("RESPONSEFORMAT not found in job");
+        } else {
```
brianmajor, line 254: Maybe "Missing required parameter: RESPONSEFORMAT", as this error will be seen by clients.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -241,31 +233,49 @@ private void doIt() {
...
+                syncOutput.setHeader("Content-Type", ZipWriter.MIME_TYPE);
+                syncOutput.setHeader("Content-Disposition", cdisp.toString());
+                this.outputStreamCounter = new ByteCountOutputStream(syncOutput.getOutputStream());
+                return new TarWriter(this.outputStreamCounter);
+
+            } else if (responseFormat.equals(TarWriter.MIME_TYPE)) {
+                cdisp.append(".tar");
+                syncOutput.setHeader("Content-Type", TarWriter.MIME_TYPE);
+                syncOutput.setHeader("Content-Disposition", cdisp.toString());
+                this.outputStreamCounter = new ByteCountOutputStream(syncOutput.getOutputStream());
```
brianmajor, line 273: The above 2 lines (and in the block above) can be brought out because they are the same in both.

brianmajor (changes requested):
Just a couple of minor comments...

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: done.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
hjeeves: Reworked this section a bit. renamed 'outputStreamCounter' to 'bcOutputStream' - the 'Counter' part was a misnomer. It's used in run() to log the amount of bytes streamed.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.commons.compress.archivers.ArchiveEntry;
+import org.apache.commons.compress.archivers.ArchiveOutputStream;
+import org.apache.log4j.Logger;
+
+public abstract class PackageWriter {
+    private static final Logger log = Logger.getLogger(PackageWriter.class);
+
+    ArchiveOutputStream aout;
+
+    public PackageWriter(OutputStream ostream) { }
```
brianmajor, line 92: It's a bit strange to have a constructor that takes a param but does nothing with it.  Why not pass the job of holding the `ostream` to this class?  (instead of each of the 2 subclasses doing it.)

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -0,0 +1,154 @@
...
+     * @param name - filename + relative path (needed so directory structure is created)
+     * @param size
+     * @param lastModifiedDate
+     * @return
+     */
+    abstract ArchiveEntry createEntry(String name, long size, Date lastModifiedDate);
+
+    public void close() throws IOException {
+        aout.finish();
+        aout.close();
```
brianmajor, line 105: What prevents `aout` from being null?  I guess there's a contract that subclasses must instantiate it?

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/ZipWriter.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public static final String MIME_TYPE = "application/zip";
+    public static final String EXTENSION = ".zip";
+
+    private OutputStream ostream;
+
+    public ZipWriter(OutputStream ostream) {
+        super(ostream);
+        this.ostream = ostream;
+        this.aout = new ZipArchiveOutputStream(ostream);;
```
brianmajor, line 92:
```suggestion
        this.aout = new ZipArchiveOutputStream(ostream);
```

brianmajor, line 92: How about the first (and only) line in the constructor is:
```super(new ZipArchiveOutputStream(ostream));```
Then PackageWriter holds on to the output stream it receives in the constructor.

**Review by brianmajor — CHANGES_REQUESTED** [review]
The arrangement of responsibility between PackageWriter and the two subclasses could use some rework.  I think my comment in the ZipWriter constructor would be the way to go, but open to opinions.

File: cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java (hunk 3 above)
Comment (hjeeves) [reply cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java:92]: Just an oversight. I don't think that ostream value even needs to be saved, as nothing is done with it after the ArchiveOutputStream is intantiated (this.aout)

File: cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java (hunk 4 above)
Comment (hjeeves) [reply cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java:105]: A null check can be added - the code would have failed long before this if aout was null.

File: cadc-pkg-server/src/main/java/org/opencadc/pkg/server/ZipWriter.java (hunk 5 above)
Comment (hjeeves) [reply cadc-pkg-server/src/main/java/org/opencadc/pkg/server/ZipWriter.java:92]: I'll see if I can make this work with setting the setLongFileMode on the tar version.
It's much tighter for zip, agreed.

File: cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java [hunk 6]
```diff
@@ -69,88 +69,34 @@
...
-    }
+    public static final String MIME_TYPE = "application/x-tar";
+    public static final String EXTENSION = ".tar";
 
-    public void close() throws IOException {
-        tout.finish();
-        tout.close();
+    public TarWriter(OutputStream ostream) {
+        super(new TarArchiveOutputStream(ostream));
+        ((TarArchiveOutputStream)super.aout).setLongFileMode(TarArchiveOutputStream.LONGFILE_POSIX);
```
pdowler, line 87: From the discussion, yeah this is a little yuck, but for a one liner it's OK. My main concern with it is that it requires `aout` to be protected rather than private in the base class.

The alternative would be to have a no-arg PackageWriter() and require (document) subclasses to assign to `super.aout` before returning from ctor. That doesn't follow the normal practice/style of Writer and OutputStream...

An alternative to documenting "assign to aout" would be to have an  `abstract ArchiveOutputStream getArchiveOutputStream()` in PackageWriter() that was called in no-arg PackageWriter()... *and* make `aout` private. My IDE complains about ctors calling overridable methods, so someone thinks that a poor programming style... but it does get you out of the issue of having to call super(...) first ickiness and allows `aout` to be private without losing clarity on what subclasses have to do.

this is all optional: the current one-extra line is OK for now

pdowler (changes requested):
abstract PackageRunner.getPackageName() is not used and should be removed

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/TarWriter.java
hjeeves: Ok. Leaving it as is. Even changing this to getAout() wouldn't escape the ugly casing going on there.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -240,32 +235,60 @@ private void doIt() {
...
+            bcOutputStream = initOutputStream(ZipWriter.MIME_TYPE, cdisp.toString());
+            return new ZipWriter(bcOutputStream);
+
+        } else if (responseFormat.equals(TarWriter.MIME_TYPE)) {
+            // Default for RESPONSEFORMAT is 'application/x-tar'
+            cdisp.append(TarWriter.EXTENSION);
+            bcOutputStream = initOutputStream(TarWriter.MIME_TYPE, cdisp.toString());
+            return new TarWriter(bcOutputStream);
+
+        } else {
```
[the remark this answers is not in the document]
hjeeves: removed else.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -240,32 +235,60 @@ private void doIt() {
...
         cdisp.append(packageName);
-        cdisp.append(".tar");
 
-        syncOutput.setHeader("Content-Type", "application/x-tar");
-        syncOutput.setHeader("Content-Disposition", cdisp.toString());
-        this.outputStreamCounter = new ByteCountOutputStream(syncOutput.getOutputStream());
+        // Initialize the writer, underlying syncoutput and output streams.
+        if (responseFormat.equals(ZipWriter.MIME_TYPE)) {
+            cdisp.append(ZipWriter.EXTENSION);
+            bcOutputStream = initOutputStream(ZipWriter.MIME_TYPE, cdisp.toString());
```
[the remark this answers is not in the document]
hjeeves: did this.* in this assignment case. thought of going through the whole file to add this. on any member variable, but thought it wasn't necessary based on this comment.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -185,13 +180,13 @@ private void doIt() {
...
 
-            // TarWriter needs an output stream instantiated.
+            // packageName should have been set to something useful in initPackage()
+            // as part of an impelmenting class
             if (StringUtil.hasText(packageName)) {
-                initOutputStream(packageName);
+                writer = initWriter();
             } else {
-                throw new IllegalArgumentException("package name not defined.");
+                throw new IllegalArgumentException("packageName not defined.");
```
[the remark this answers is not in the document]
hjeeves: reworked this, and changed the exception.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -97,19 +91,20 @@
 /**
  * Abstract class PackageRunner: a JobRunner implementation that provides core
  * functionality for building CADC package files (zip or tar files, for example.)
+ * To select a package type, RESPONSEFORMAT must be provided as a job parameter.
```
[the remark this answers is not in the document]
hjeeves: reworded, removed 'must'

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
```diff
@@ -3,7 +3,7 @@
  *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
  **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
  *
- *  (c) 2021.                            (c) 2021.
+ *  (c) 2022.                            (c) 2022.
```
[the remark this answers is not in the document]
hjeeves: put date back to what it was.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageRunner.java
```diff
@@ -97,19 +91,20 @@
...
  * Job state management and error reporting back to the Job are provided.
  */
 public abstract class PackageRunner implements JobRunner {
     private static final Logger log = Logger.getLogger(PackageRunner.class);
 
-    private JobUpdater jobUpdater;
     private SyncOutput syncOutput;
-    private ByteCountOutputStream outputStreamCounter;
+    private ByteCountOutputStream bcOutputStream;
```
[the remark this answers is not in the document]
hjeeves: put JobUpdater back where it was

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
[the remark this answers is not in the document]
hjeeves: as to the 'getPackageName()' removal - I'll do it here, and there'll be a small change in caom2-pkg-server needed to adapt to this when it's next updated. (it was a remnant of an idea that should be removed.)

pdowler (approved):

PR #81, opened 2022-04-14
CADC-11262: add file:// URL support in PackageWriter

hjeeves:
Updated Tar and ZipWriterTests changed to use test/resources files instead of relying on known good URLs.

Has been tested using cavern package integration tests.


cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -109,49 +115,118 @@ public void close() throws IOException {
...
+
+    /**
+     * Set up header for ArchiveEntry in the current ArchiveOutputStream.
+     * @param relativePath - relative path of item in the package file
+     * @param contentLength - (long) - bytes to be written
+     * @param lastModified - (Date) - last modified date of item
+     * @return
+     * @throws IOException
+     */
+    private boolean setArchiveEntryHeader(String relativePath, long contentLength, Date lastModified)
```
brianmajor, line 166: Is the return type 'boolean' here necessary?  It only returns 'true' (never false) so that is the same as saying "this method completed", so don't think it adds value.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -109,49 +115,118 @@ public void close() throws IOException {
...
+     * @param fileURL - URL of file item being written to
+     * @param relativePath - relative path of file inside package beingw written.
+     * @throws IOException
+     */
+    private void writeFileContentEntry(URL fileURL, String relativePath) throws IOException {
+        Path filePath = FileSystems.getDefault().getPath(fileURL.getPath());
+        log.debug("filePath from fileURL: " + filePath.toString() + ": " + fileURL.toString());
+
+        long contentLength = Files.size(filePath);
+        FileTime lastMod = (FileTime)Files.getAttribute(filePath, "lastModifiedTime");
```
brianmajor, line 192: There is a 'getLastModifiedTime' method that's probably better to use.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -88,6 +93,7 @@ public abstract class PackageWriter {
     private static final Logger log = Logger.getLogger(PackageWriter.class);
 
     ArchiveOutputStream aout;
+    boolean openEntry;
```
brianmajor, line 96: Having this attribute outside the scope of the method in which it is used makes the class not thread safe.

brianmajor (changes requested):
The declaration of 'openEntry' should be put back to where it was.  The two methods that stream should keep track of their state separately (set their local openEntry to true after attributes set) and return boolean values themselves.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
hjeeves: It is relevant if the step for setting up the archive metadata fails, as compared to when writing the body of the archive entry fails.  - 'openEntry = true' means that the archive metadata write was successful. 'openEntry=false' means it wasn't.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
hjeeves: ok

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
hjeeves: ok

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
brianmajor: But since this cannot return false it should not return anything.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -109,47 +114,125 @@ public void close() throws IOException {
...
+        // get information common to all entries
+        long contentLength = get.getContentLength();
+        Date lastModified = get.getLastModified();
+        boolean openEntry = false;
 
+        try {
             // headers for entry have been written, body has not,
             // so consider this entry 'open'
-            openEntry = true;
+            openEntry = setArchiveEntryHeader(relativePath, contentLength, lastModified);
```
brianmajor, line 223: Just set `openEntry = true` when setArchiveEntryHeader completes.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -109,47 +114,125 @@ public void close() throws IOException {
...
         try {
-            // HEAD to get entry metadata
-            // Implementations of PackageRunner that build the PackageItem list
-            // should ensure that files exist before submitting as part of the
-            // Iterator<PackageItem>
-            URL packageURL = packageItem.getURL();
-            HttpGet get = new HttpGet(packageURL, true);
-            get.prepare();
+            // This gets set to true unless setArchiveEntryHeader throws an error
+            openEntry = setArchiveEntryHeader(relativePath, contentLength, lastModified);
```
brianmajor, line 187: Just set `openEntry = true` when setArchiveEntryHeader completes.

brianmajor (changes requested):

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
hjeeves: ok

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
hjeeves: ok

brianmajor (approved):

PR #86, opened 2022-07-26
Cone Search library

at88mph:


cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
```diff
@@ -0,0 +1,236 @@
...
+    private static final int MID_VERB_VALUE = 2;
+    private static final int MAX_VERB_VALUE = 3;
+    private static final Logger LOGGER = LogManager.getLogger(TAPQueryGenerator.class);
+
+    protected final String catalog;
+    protected final String lowVerbositySelectList;
+    protected final String midVerbositySelectList;
+    protected final String highVerbositySelectList;
+
+    public TAPQueryGenerator(final String catalog, final String lowVerbositySelectList,
```
pdowler, line 100: The position column also must be a ctor arg since it is tightly coupled with the table name. `catalog` is pretty generic so I would use `tableName` and `posColumnName`.

The verb-dependent select lists I see as a pretty fragile way to pick columns, but I can't think of anything bulletproof so I guess simplest is the right choice now. I can imagine too many variations and they would all be complex :-)

This generator is a different style from the sia2/AdqlQueryGenerator; that class takes the input param map in the ctor and returns the TAP params from a no-arg `getParameterMap(). I can see you looked at that because both of them have the same incorrect comment mentioning the REQUEST param (which no longer exists in TAP). The sia2 style makes a single use generator, while this allows calling code to generate different queries with different inputs; there are no use cases for the latter since a cone search request is inherently single query.

Finally, the ctor should check all args for null and throw IllegalArgumentException if necessary.

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Map;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+/**
+ * Class to produce an ADQL query from a set of Parameters
+ */
+public class TAPQueryGenerator extends CommonParamValidator {
```
pdowler, line 87: As I mentioned in another comment, I expected something like `ConeParamValidator extends CommonParamValidator` and that the query generator would use it (rather than be the thing extending). It separates the two things and allows the ConeParamValidator to be used without the query generator. Given that SCS has 3 params that map to one `Circle` constraint, I would probably have a method like 
```
public Circle ConeParamValidator.validateCone(Map<String, List<String>> params)
```
to encapsulate/manage that old style.

So, you should end up with ConeParamValidator and TAPQueryGenerator (ugh, acronyms in class names).

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param positionColumnName The name of the positional column to compare against the cone's position,
+     *                           preferably spatially indexed.
+     * @return map of parameter names and values
+     */
+    public Map<String, Object> getParameterMap(final String positionColumnName,
+                                               final Map<String, List<String>> parameters) {
+        final Map<String, Object> queryParameterMap = new HashMap<>();
+
+        // Obtain and, if necessary, provide a default RESPONSEFORMAT.
+        final String requestedResponseFormat = getFirstParameter(ParameterNames.RESPONSEFORMAT.name(), parameters,
```
pdowler, line 120: ConeSearch allows for output format other than VOTable? 

If so, use `CommonParamValidator.getResponseFormat(...)`

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
```diff
@@ -0,0 +1,236 @@
...
+                                                                 VOTableWriter.CONTENT_TYPE);
+        queryParameterMap.put(ParameterNames.RESPONSEFORMAT.name(), requestedResponseFormat);
+
+        // Obtain and validate the VERB (verbosity) output.
+        final String requestedOutputVerbosity = getFirstParameter(ParameterNames.VERB.name(), parameters,
+                                                                  Integer.toString(MID_VERB_VALUE));
+        final NumberParameterValidator outputVerbosityValidator =
+                new NumberParameterValidator(false, TAPQueryGenerator.MIN_VERB_VALUE,
```
pdowler, line 128: rather than a separate NumberParameterValidator, just use `CommonParamValidator.validateInteger(...)`. You'll have to deal with getting the first value from the list here (multi-valued params are the norm now so not generically supported in CommonParamValidator)

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
```diff
@@ -0,0 +1,236 @@
...
+                                             TAPQueryGenerator.MAX_VERB_VALUE, TAPQueryGenerator.MID_VERB_VALUE);
+        final int outputVerbosity = outputVerbosityValidator.validate(requestedOutputVerbosity);
+
+        // Obtain and validate the MAXREC value with a default if necessary.
+        final String requestedMaxRecords = getFirstParameter(ParameterNames.MAXREC.name(), parameters,
```
pdowler, line 133: MAXREC should be added to the CommonParamValidator, eg
```
public Integer getMaxRec(Map<>, Integer defaultValue, Integer maxValue)
```
where both integer args could be null (server-side choice). There is a MaxRecValidator in cadc-dali that is used in TAP, but it currently requires a Job so isn't usable here, but the code in it's validate() method is easily adapted.

pdowler (changes requested):

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
at88mph: It does as of 1.1:
https://www.ivoa.net/documents/ConeSearch/20200828/WD-ConeSearch-1.1-20200828.html#tth_sEc2.1.6

at88mph:
Review work submitted.  They're a little lengthy, and now include changes to the `CommonParamValidator` in `cadc-dali`.

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
at88mph: This is done.

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
at88mph: This is done.

cadc-conesearch/src/main/java/org/opencadc/conesearch/TAPQueryGenerator.java
at88mph: This is done.

cadc-conesearch/src/main/java/org/opencadc/conesearch/ParameterNames.java
```diff
@@ -0,0 +1,84 @@
...
+ *  <http://www.gnu.org/licenses/>.      pas le cas, consultez :
+ *                                       <http://www.gnu.org/licenses/>.
+ *
+ *
+ ************************************************************************
+ */
+
+package org.opencadc.conesearch;
+
+public enum ParameterNames {
```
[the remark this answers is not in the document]
at88mph: Done.

cadc-conesearch/README.md
```diff
@@ -0,0 +1,68 @@
...
+positionColumnName = my_spatial_indexed_column  # Used to compare to the Cone's Center position.  Can also be a dynamic spatial object e.g. POINT('ICRS', RA_Source, DEC_Source)
+lowVerbositySelectList = obs_id, ra, dec        # Columns for VERB=1
+midVerbositySelectList = obs_id, ra, dec, s_region, frequency, bandwidth  # Columns for VERB=2
+highVerbositySelectList = *  # Columns for VERB=3
+```
+
+Capture the Cone Search parameters and translate to ADQL to be sent to a TAP service.
+
+```java
+final ConeSearchConfig coneSearchConfig = new ConeSearchConfig();  // Ensure the cadc-conesearch.properties exists.
```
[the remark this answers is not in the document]
at88mph: Done.

cadc-conesearch/README.md
```diff
@@ -0,0 +1,68 @@
...
+    mavenCentral()
+    mavenLocal()
+}
+
+dependencies {
+    implementation 'org.opencadc:cadc-conesearch:1.0.0'
+}
+```
+
+`${HOME}/config/cadc-conesearch.properties`:
```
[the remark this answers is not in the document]
at88mph: Right, done.

PR #91, opened 2022-10-11
Ngvs cutout fix

at88mph:
Fix to use uncompressed dimensions (`ZNAXISn`) for slicing.

### DIFF
--- a/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/ShapeCutout.java
+++ b/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/ShapeCutout.java
@@ -72,8 +72,10 @@ import ca.nrc.cadc.dali.Shape;
 
 import java.io.Serializable;
 
+import nom.tam.fits.FitsException;
 import nom.tam.fits.Header;
 import nom.tam.fits.HeaderCardException;
+import nom.tam.fits.header.Compression;
 import nom.tam.fits.header.Standard;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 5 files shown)


cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/ShapeCutout.java
```diff
@@ -94,6 +98,30 @@ public ShapeCutout(FITSHeaderWCSKeywords fitsHeaderWCSKeywords) {
...
+        final int nAxis = this.fitsHeaderWCSKeywords.getIntValue(Compression.ZNAXIS.key(),
+                                                                 this.fitsHeaderWCSKeywords.getIntValue(
+                                                                         Standard.NAXIS.key()));
+
+        if (nAxis == 0) {
+            return new int[0];
+        }
+
+        final int[] axes = new int[nAxis];
+        for (int i = 1; i <= nAxis; i++) {
```
pdowler, line 116: Blindly using ZNAXISi or NAXISi here seems dangerous. If you actually used a value from ZNAXIS above, you should only use ZNAXISi here (and similarily for NAXIS and NAXISi). It could be that the problem only shows up in incorrect fits headers, but it would go unnoticed and probably make it harder to figure out if this code isn't more careful.

pdowler (changes requested):

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/ShapeCutout.java
at88mph: OK.  I've clarified when it gets used.  Thanks.

pdowler (approved):

PR #93, opened 2023-02-09
cadc-datalink: add local_semantics, remove readable

pdowler:
add basic serialization in votable test


pdowler:
oops, requires cadc-registry-1.6.2

PR #96, opened 2023-12-01
CADC-12563 support to add directories and symbolic links to packages

jburke-cadc:


pdowler (changes requested):
I see mention of symbolic links in the PR title but superficially nothing in the code. Are they not supported in both tar and zip?

Is polymorphism really the best approach here? It isn't the style of java IO normally, for example with `java.io.File` and it's `isDirectory()` and `isFile()`. With File, regular file vs link requires comparing absolute path and canonical path (subtle). Or this in java.nio:

```
PosixFileAttributes attrs = Files.readAttributes(p,
                PosixFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
        if (attrs.isDirectory()) {
            ...
        } else if (attrs.isRegularFile()) {
            ...
        } else if (attrs.isSymbolicLink()) {
            ...
        }
```
I think something simpler like multiple constructors in PackageItem and some boolean methods as above (in nio) would suffice (exactly which ones depending on the answer to the symbolic link question). 

general principle: I've come to realise that one should only use polymorphism where it's really needed and where it can remove the need for conditionals and treat different subtypes the same way. Use of `instanceof` in the same library is a strong hint that this case doesn't really fit the principle (it isn't right vs wrong, it's complexity vs benefit).

pdowler:
let's see:
PackageItem(String relativePath) // directory
PackageItem(String relativePath, URL content) // file
PackageItem(String relativePath, String linkTarget) // symlink

As long as you check and do not allow any null constructor args, it should be unambiguous and you can implement the same methods as in the posix attrs API.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
```diff
@@ -73,42 +73,119 @@
...
+     *
+     * @return URL to the content for a file, null if the item is a directory or symbolic link.
+     */
+    public URL getContent() {
+        return this.content;
     }
 
-    public String getRelativePath() {
-        return relativePath;
+
```
pdowler, line 163: relativePath is never null via ctor checks,  so those checks below are not needed

otherwise, isFile only needs `content != null` and isSymbolicLink only needs `linkTarget != null`

style: I prefer to use the `this.` qualifier when assigning to a member variable (LHS of =), but not when using it. It makes it more clear when "state" is being modified.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
```diff
@@ -131,110 +133,102 @@ public void write(PackageItem packageItem) throws IOException, InterruptedExcept
...
-            if (openEntry) {
-                // header was written but body was not, try to close gracefully
-                aout.closeArchiveEntry();
-            }
-        }
+        // put archive entry to stream
+        archiveOutputStream.putArchiveEntry(archiveEntry);
 
+        // copy the file to archive output stream
+        Files.copy(filePath, archiveOutputStream);
```
pdowler, line 181: I'm pretty sure MultiBufferIO.copy is faster than this builtin function because it reads and writes in separate threads. It can make a difference when one of the streams (usually the client stream) is slower than the other.

You can just use `fileURL.openStream()` in this case.

cadc-pkg-server/src/test/java/org/opencadc/pkg/server/TarWriterTest.java
```diff
@@ -94,28 +91,45 @@ public class TarWriterTest {
...
+            String link1Path = "some/path/link2SymbolCanada.gif";
+            PackageItem link1 = new PackageItem(link1Path, file2Path);
+            log.debug(link1);
+
+            List<PackageItem> packageContents = new ArrayList<>();
+            packageContents.add(dir1);
+            packageContents.add(dir2);
+            packageContents.add(file1);
+            packageContents.add(file2);
+            packageContents.add(link1);
```
pdowler, line 129: seems like the package content above is the same as in the zip test... maybe have a single method to generate the test content.

also, why does the test write tom a file and to a byte array? I would just write/read the file and leave it behind for manual inspection

pdowler (changes requested):

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageItem.java
jburke-cadc: Updated for first two comments. I think I've heard the third before and will try to better follow it.

cadc-pkg-server/src/main/java/org/opencadc/pkg/server/PackageWriter.java
jburke-cadc: MultiBufferIO.copy is already used for HTTP urls, updated to use it for file urls as well.

cadc-pkg-server/src/test/java/org/opencadc/pkg/server/TarWriterTest.java
jburke-cadc: Moved the zip and tar writer tests into a single PackageWriterTest class. A single method to create the PackageItem's to be archived, and writing/reading the archive without checking for individual files, really shrunk down the size of the test methods, so put it into a single class.

pdowler (approved):

PR #106, opened 2024-12-03
Support Parquet as Response format.

aratikakadiya:


cadc-dali/build.gradle
```diff
@@ -28,6 +28,11 @@ dependencies {
     implementation 'org.opencadc:cadc-util:[1.6,)'
     implementation 'org.opencadc:cadc-uws:[1.0.4,)'
 
+    implementation 'org.apache.parquet:parquet-avro:1.15.0'
+    implementation 'org.apache.parquet:parquet-hadoop:1.15.0'
+    implementation 'org.apache.hadoop:hadoop-common:3.3.6'
+    implementation 'org.apache.hadoop:hadoop-mapreduce-client-core:3.3.6'
```
pdowler, line 34: are all these libs needed to read/write parquet? the last one seems out of scope...

in general, I prefer to declare dependencies a little more flexibly, eg

feeling lucky/lazy:
```
implementation 'org.apache.parquet:parquet-avro:[1.15.0,)'
```
slightly less risky:
```
implementation 'org.apache.parquet:parquet-avro:[1.15.0,2.0)'
```
even more risk averse:
```
implementation 'org.apache.parquet:parquet-avro:[1.15.0,1.16)'
```
Probably "slightly less risky" is a good general choice unless we know otherwise.

pdowler (changes requested):
The CI is failing checkstyle; you vcan run it locally with
```
gradle checkstyleMain
```

Some general comments on the code below...

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetWriter.java
```diff
@@ -0,0 +1,205 @@
...
+
+            TableData tableData = resource.getTable().getTableData();
+            resource.getTable().setTableData(null);
+
+            StringWriter stringWriter = new StringWriter();
+            VOTableWriter votableWriter = new VOTableWriter();
+            votableWriter.write(voTableDocument, stringWriter, maxRec);
+
+            Map<String, String> customMetaData = new HashMap<>();
+            customMetaData.put("votable", stringWriter.toString());
```
pdowler, line 81: The note on VOTable in parquet specifies two specific keys to use here.

https://github.com/ivoa/voparquet

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetWriter.java
```diff
@@ -0,0 +1,205 @@
...
+                    log.debug("Total Records generated= " + (recordCount - 1));
+                }
+            } catch (Exception e) {
+                throw new IOException("error while writing", e);
+            }
+        }
+        out.close();
+    }
+
+    @Override
```
pdowler, line 120: I think we should remove the two TableWriter.write(..., Writer, ...) methods from the interface. It's clearly a bad idea since it only applies to text-based formats. Then these dummy methods could be removed from here.

It would be OK to keep those methods as convenience in the inherently text (AsciiTableWriter and VOTableWriter) but would need to remove the `@Override`. 

---

I am less sure what we should do about `write(Throwable, OutputStream)` here... I guess since getErrorContentType() says "text/plain" it should output a plain text error message (for now at least). There are reasons, but complicated and from othewr IVOA standards.

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        if (recordCount == 0) {
+            throw new RuntimeException("NO Records Read");
+        }
+        return new TableShape(recordCount, columnCount);
+    }
+
+    private static InputFile getInputFile(InputStream inputStream) throws IOException {
+
+        // Read the entire InputStream into a byte array
```
pdowler, line 71: OK for now, but we can't use this class with any real (large) content without considerable thought.

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -0,0 +1,190 @@
...
+
+        InputFile inputFile = getInputFile(inputStream);
+        int recordCount = 0;
+        int columnCount = 0;
+
+        ParquetMetadata metadata = ParquetFileReader.open(inputFile).getFooter();
+        MessageType schema = metadata.getFileMetaData().getSchema();
+        columnCount = schema.getFieldCount();
+
+        String votable = metadata.getFileMetaData().getKeyValueMetaData().get("votable");
```
pdowler, line 40: You could read this xml string using VOTableReader and then add the VOTableDocument object to the TableShape object that is returned... we could then test that metadata round trip is correct.

I feel like the future of this read() method will be to return a class that holds various things... and the rich metadata will be one of those things. I'm not sure what form the returned data might take, so don't think we should do anything for that just yet.

pdowler (changes requested):

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/DynamicSchemaGenerator.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        // Define the Avro record schema with the fields
+        Schema schema = Schema.createRecord("Record", null, null, Boolean.FALSE);
+        schema.setFields(fields);
+        log.debug("Schema Generated Successfully : " + schema);
+        return schema;
+    }
+
+    private static Schema getAvroFieldType(String sqlType) {
```
pdowler, line 40: First time looking at this file... everything else looks OK.

I think this method needs to look at more of the VOTableField metadata. The complete datatype is defined by `datatype`, `arraysize`, and `xtype`. The latter two are optional and if null it means the data is a single primitive.

For "char" that actually means a single char; a string is technically `datatype="char" arraysize="X"` (for some value of X allowed by the VOTable spec). Maybe those are the same to avro. 

A few xtypes like "uri" and "uuid" you could probably ignore for now, but we should try to do "timestamp" correctly. We can also ignore geometry for now (point, circle, polygon, etc) but will need to address if before release.

So the changes: change method signature to `getAvroFieldType(VOTableField)`, check the char vs string correctness, and fix the timestamp (it's a char array with an xtype):

`datatype="char" arraysize="X" xtype="timestamp"`

In that column, the value objects should be java.util.Date and I don't know if avro will format that correctly in a standard way or if we need to format it in a VO-compatible way. Any idea?

pdowler (changes requested):

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/DynamicSchemaGenerator.java
pdowler: iirc, in miost of our TAP services timestamp columns have this metadata:
```
datatype="char" arraysize="23*" xtype="timestamp"
```
that arraysize means variable length with a max of 23 characters... just noting that you probably do not need to look at the value of arraysize (at least for this one), just the xtype.

---

In future, if you need to know the arraysize it "is complicated" because length and variable-size are both embedded in there, and in a few cases there are even multi-dimensional arrays. An example of that is the caom2.Plane.energy_bounds_samples column. The datatype will be Interval[] or List<Interval> and in VOTable that looks like
```
datatype="double" arraysize="2x*" xtype="interval"
```
where 2x* means a 2 x var-length array aka an array of 2-value intervals.
We only have a few cases of multidimensional:

pdowler: from the discussion around https://github.com/ivoa/voparquet/issues/10, I think we should aim for using the parquet logical types where applicable (e.g. make the "best" parquet file). 

In the case of timestamp that means modifying the votable FIELD metadata to match and be DALI compliant, so `datatype="long"` and remove arraysize and xtype.

For uuid, let's do the opposite: logical type STRING and keep the votable metadata as-is. The  we will be able to point others to an example of both approaches for comparison.

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetWriter.java
```diff
@@ -0,0 +1,306 @@
...
+
+    private static void updateVOTable(VOTableResource voTableResource) {
+        List<VOTableField> fields = voTableResource.getTable().getFields();
+
+        // Convert timestamp fields to long to ensure inconsistencies between avro schema field and votable metadata field
+        for (int i = 0; i < fields.size(); i++) {
+            VOTableField field = fields.get(i);
+
+            if ("timestamp".equals(field.xtype)) {
+                fields.set(i, new VOTableField(field.getName(), "long"));
```
pdowler, line 201: Yes, the VOTableField datatype is immutable so you have to create a new one.. but there are a bunch of optional metadata in the source field as well (aka public fields that can be null): some of that is still applicable and some is not.

In the case of timestamp (aside from the datatype,arraysize,xtype change): the units could be set explicitly to `msec` (I think - you will need to check the VOUnits standard) but I think everything else is still applicable.

In the case of short below, all the optional metadata is still applicable.

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetWriter.java
```diff
@@ -0,0 +1,306 @@
...
+
+    private static void handleArrays(Schema.Field field, String xtype, GenericRecord record, Object data) {
+        if (xtype == null) {
+            record.put(field.name(), data);
+        } else {
+            handleXTypeArrayData(record, field.name(), xtype, data);
+        }
+    }
+
+    private static void handleXTypeArrayData(GenericRecord record, String columnName, String xtype, Object data) {
```
pdowler, line 265: Needing to use this much reflection is a bad sign and I think this could be improved, but not now: I have some ideas to improve/simplify Interval (in my CAOM 2.5 branch) and that might help simplify this. Nothing to do for now - this will work/fail correctly as-is.

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetWriter.java
```diff
@@ -0,0 +1,306 @@
...
+    private static int saveRecords(Long maxRec, TableData tableData, Schema schema, org.apache.parquet.hadoop.ParquetWriter<GenericRecord> writer)
+            throws IOException {
+        Iterator<List<Object>> iterator = tableData.iterator();
+        int recordCount = 1;
+
+        while (iterator.hasNext() && recordCount <= maxRec) {
+            GenericRecord record = new GenericData.Record(schema);
+            List<Object> rowData = iterator.next();
+
+            for (Schema.Field field : schema.getFields()) {
```
pdowler, line 234: I would prefer this `for` loop be over the rowData list so this is more clearly "loop over the data", but then you need tom be able to get the field from the schema by position... don't know if that's supported.

pdowler (changes requested):

pdowler (approved):

PR #110, opened 2025-02-18
CADC-13577 add support to the VOTable root for an INFO element with a ID attribute

jburke-cadc:


cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/votable/VOTableDocument.java
```diff
@@ -77,13 +77,14 @@
...
-    private List<VOTableResource> resources = new ArrayList<VOTableResource>();
+    private List<VOTableResource> resources = new ArrayList<>();
+    private List<VOTableInfo> infos = new ArrayList<>();
 
     public VOTableDocument() {
     }
 
     public List<VOTableResource> getResources() {
-        return resources;
+        return this.resources;
```
pdowler, line 87: the explicit `this.` is not really needed here. It's useful when assigning values to member variables (because it allows argument and assignment to use the same name in ctor and set methods) but not otherwise.... especially in a simple get method where there is no other possibility.

ditto below in getInfos()

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/votable/VOTableWriter.java
```diff
@@ -287,19 +287,6 @@ protected void writeImpl(VOTableDocument votable, Writer writer, Long maxrec)
                 Element table = new Element("TABLE", namespace);
                 resource.addContent(table);
 
-                // Create the INFO element and add to the RESOURCE element.
```
pdowler, line 290: The more compact version using createInfo() is fine, but why move all the INFO elements to after the params and table data? 

There is a use for trailing INFO elements (error and overflow), but that is separate and explicit

pdowler (changes requested):

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/votable/VOTableDocument.java
jburke-cadc: Updated

cadc-dali/src/main/java/ca/nrc/cadc/dali/tables/votable/VOTableWriter.java
jburke-cadc: I don't remember why I moved it, maybe because I added the same statements to the VOTable root, so I put them side by side?

pdowler (approved):

PR #111, opened 2025-02-18
CADC-13577 added new ServiceDescriptorTemplate class and unit tests

jburke-cadc:


cadc-datalink/src/test/java/org/opencadc/datalink/ServiceDescriptorTemplateTest.java
```diff
@@ -0,0 +1,228 @@
...
+        try {
+            new ServiceDescriptorTemplate("test", template);
+        } catch (IllegalArgumentException e) {
+            Assert.fail("Expected no exception for valid template");
+        }
+    }
+
+    @Test
+    public void testInvalidIdRefTemplate() throws Exception {
+
```
pdowler, line 156: All of these fail because of the exception from the `read` method; that error message is from the XML parser.

So most of these tests are just more elaborate than necessary and the code to re-validate never actually does anything useful. That's kind of the point of using xml: a lot of validation is built in (sure, error messages can be bad but that's a different problem). One test with a very simple mismatched ID/IDREF usage is sufficient, so probably just keep the mismatched-id-ref-template but remove the matching one and just have a simple mismatched one.

I'll make some comments in the valid-template about minimal content in there, but it should apply to all the test files.

cadc-datalink/src/main/java/org/opencadc/datalink/ServiceDescriptorTemplate.java
```diff
@@ -0,0 +1,204 @@
...
+    private List<String> parseIdentifiers(String template) {
+        VOTableReader reader = new VOTableReader();
+        VOTableDocument votable;
+        try {
+            votable = reader.read(template);
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Error reading VOTable template: " + e.getMessage());
+        }
+
+        // List of ID's from the INFO elements
```
pdowler, line 166: This and the check below (Validate that all ID's have a corresponding ref) are useless because the XML parser already checked that and threw the exception from read(). As noted in the test, you already confirmed this to be the case from the message always being that "ID/IDREF binding" thing

cadc-datalink/src/main/java/org/opencadc/datalink/ServiceDescriptorTemplate.java
```diff
@@ -0,0 +1,204 @@
...
+        if (infoIDs.isEmpty()) {
+            throw new IllegalArgumentException("template must contain one or more info elements "
+                    + "with an ID attribute in the votable root.");
+        }
+
+        List<VOTableResource> resources = votable.getResources();
+        if (resources.isEmpty()) {
+            throw new IllegalArgumentException("template must contain at least one resource element.");
+        }
+        VOTableResource resource = resources.get(0);
```
pdowler, line 180: Above says "must be at least one" but then you only look at one here. There can be multiple... naively someone might do that if they want multiple service descriptors with the same IDREF (or set of IDREF). For example, there are several services one can invoke using the CAOM publisherID and they could all be in one template... but it is an uncommon usage and we could (for now) limit to a single ServiceDescriptor resource in the template.

So maybe for now change the check above to `resources.size() == 1`

cosmetic: when writing error messages in exceptions, I prefer not trying to make them sentences, so don't capitalise unless it is for emphasis (eg BUG: ...) and don't include pointless punctation at the end.... here I see non-capitalised (ok) and a period (indicating a sentence end: ugh)... it is easier to just write "phrases" or semi-structured messages (sometime use colon, eg `BUG: unexpected situation...` or `invalid URI: ... reason: ...`

pdowler (changes requested):

cadc-datalink/src/test/java/org/opencadc/datalink/ServiceDescriptorTemplateTest.java
jburke-cadc: Cleaned up the unit tests to check only what is thrown outside of the schema validation.

cadc-datalink/src/main/java/org/opencadc/datalink/ServiceDescriptorTemplate.java
jburke-cadc: Removed redundant checks in the class.

cadc-datalink/src/test/resources/valid-template.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pdowler, line 1: I forgot to comment on the valid example. I'd prefer to use generic values rather than our current service values.

Change the value of the INFO element to `placeholder: will be removed`

Only one of resourceIdentifier and accessURL is needed, so maybe 
- remove the resourceIdentifier
- change the accessURL to `https://example.net/datalink/links` (change arraysize to `*`)

Keep the standardID as-is.

Let's leave the ID and ref as-is as well; promote CAOM a bit :-)

Simplify and make other tets files generic as well.

pdowler (changes requested):

pdowler (approved):

PR #121, opened 2025-07-23
Parquet upload

aratikakadiya:


cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -69,199 +69,165 @@
...
-        log.debug("VOTable: " + votable);
+        InputFile inputFile = new ParquetInputFile((ByteArrayInputStream) inputStream);
 
-        VOTableReader voTableReader = new VOTableReader();
-        VOTableDocument voTableDocument = voTableReader.read(votable);
+        VOTableDocument voTableDocument;
 
-        try (org.apache.parquet.hadoop.ParquetReader<GenericRecord> reader = AvroParquetReader.<GenericRecord>builder(inputFile).build()) {
-            GenericRecord record;
+        try (ParquetFileReader reader = ParquetFileReader.open(inputFile)) {
```
pdowler, line 125: A lot of the changes look like they are from removing avro, so this is only a cursory look.

The cast above is not going to work in general... but it implies some assumptions about the source (and size) of the parquet data. The argument InputStream is often a network stream (essentially the input from a PUT or POST inside a servlet). I put the comment here because the crux is the first line below: 
```
ParquetMetadata metadata = reader.getFooter();
```
I looked briefly at ParquetInputFile and see that it appears we really need to seek to the end of the stream to get this footer with additional metadata. In the general case (`POST {massive parquet file} https://...`) we cannot seek to get the footer and rewind to get the data later.

So I guess the real questions are: 
- should the method signature of `read(??)` be something more specific than InputStream?
- in the general case, do we have to read and cache the parquet file somewhere we can seek?
- who is responsible for providing a seekable (random access) input data?

aratikakadiya: Great questions! I never drafted my understanding before today.

1. Sure, I am already stuck at this part as I assumed that it is going to be ByteArrayInputStream(Got this idea from our Int Tests - but its really not). Right now I am getting CloseWrapper which is a FilterInputStream. So having a middle layer which can give a stream which is seekable, that will solve the problem. I did not really know that I can expect some specific InputStream in the reader thinking its the standard.
2. Yes, In general case, if we dont have seekable stream, we have to either cache everything in the memory or have temp file. But if a stream have internal buffer, we can use it using reflection. ByteArrayInputStream has it and I have used it in ParquetInputFile class. 
3. I am not really sure about it actually whether it should be reader or whoever is calling it.

This is what I understand the execution of the reader's strategy - This might help us figuring out if seeking to the footer first is okay or not.

- ParquetFileReader uses the prepared InputFile, which wraps the ByteArrayInputStream's internal buffer.
- It knows that the last 4 bytes are PAR1 (magic number) and the 4 bytes before that are the footer length.
- Using the footer length, it calculates the starting position of the footer and calls the seek method to move to that location.
- It reads the footer into an object which we access when needed. 
- Multiple seekableInputStream is created to read the footer, row group,... but all of them uses same buffer. (Should have no issues as the exact location of footer, row groups are retrieved from the footer)

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/io/HttpRandomAccessSource.java
```diff
@@ -0,0 +1,67 @@
...
+            throw new IOException("Seek position out of bounds");
+        }
+        this.position = position;
+    }
+
+    @Override
+    public int read(byte[] buffer, int offset, int length) throws IOException {
+        long end = position + length - 1;
+
+        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
```
pdowler, line 30: HTTP code should use the classes in cadc-util (ca.nrc.cadc.net), specifically HttpGet. That will automatically perform all kids of authentication using credentials in the current Subject and do a lot of error handling. For the usage here, it would look something like
```
HttpGet get = new ...
get.prepare();
InputStream in = get.getInputStream();
```
where `prepare()` throws a variety of exceptions (eg it maps various http status codes to exceptions; the mapping is consistent with exception -> code mapping in the `cadc-rest` library). 

---

Having said that, this class will currently open a new connection for each (usually pretty small) buffer of data to be read. We haven't put much effort into making HttpGet use http keep-alive so while this will work, it is going to perform very poorly in practice.

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/io/HttpRandomAccessSource.java
```diff
@@ -0,0 +1,67 @@
...
+        return contentLength;
+    }
+
+    @Override
+    public void close() {
+        // No persistent connection to close
+    }
+
+    private long fetchContentLength(URL url) throws IOException {
+        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
```
pdowler, line 56: ca.nrc.cadc.net.HttpGet can perform head requests

also, this head request should check the http headers to see if the server accepts range requests:
```
accept-ranges: bytes
```
I can't recall what happens in a GET request with range header if the server does not support ranges.

aside: I see that our `minoc` services to not include `accept-ranges` even though they do; I'll create a bug report for that as it will matter.

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -69,199 +69,248 @@
...
-        String votable = metadata.getFileMetaData().getKeyValueMetaData().get("IVOA.VOTable-Parquet.content");
-        log.debug("VOTable: " + votable);
+    public VOTableDocument read(RandomAccessSource randomAccessSource) throws IOException {
+        RandomAccessSeekableInputStream parquetInputFile = new RandomAccessSeekableInputStream(randomAccessSource);
+        return read(parquetInputFile);
+    }
 
-        VOTableReader voTableReader = new VOTableReader();
-        VOTableDocument voTableDocument = voTableReader.read(votable);
+    public VOTableDocument read(InputStream inputStream) throws IOException {
```
pdowler, line 138: The `read` methods here should have javadocs and this one must indicate that it reads the stream fully and stores it in a local temporary file before returning.

I do not think the byte limit and ByteCountInputStream should be added here. The actual limit will depend on context so the caller can/should wrap the input stream if necessar. For example, the BasicUploadManager in cadc-tap-server already adds a ByteCountInputStream limitation.

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -69,199 +69,248 @@
...
+            String fileName = "parquet-" + UUID.randomUUID();
+            tempFile = File.createTempFile(fileName, ".parquet");
 
-                // TODO: This is a temporary implementation to read the data
-                Map<String, Object> rowData = new HashMap<>();
-                for (Type field : schema.getFields()) {
-                    String fieldName = field.getName();
-                    rowData.put(fieldName, record.get(fieldName));
+            try (InputStream byteCountInputStream = new ByteCountInputStream(inputStream, BYTE_LIMIT);
+                OutputStream out = new FileOutputStream(tempFile)) {
```
pdowler, line 147: There are lots of ways to do this inputstream -> outputstream copy loop. There are some convenience methods in java.nio and for higher performance there is a class in ca.nrc.cadc.io named `MultiBufferIO` than implements double and triple buffer copy using one additional thread. I would recommend MultiBufferIO here.

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -69,199 +69,272 @@
...
-                Map<String, Object> rowData = new HashMap<>();
-                for (Type field : schema.getFields()) {
-                    String fieldName = field.getName();
-                    rowData.put(fieldName, record.get(fieldName));
+            log.debug("File" + fileName + " prepared successfully for reading");
+
+            RandomAccessFile raf = new RandomAccessFile(tempFile, "r");
+            ParquetInputFile parquetInputFile = new ParquetInputFile(raf);
+            return read(parquetInputFile);
+        } catch (IOException e) {
```
pdowler, line 184: I think this should not catch IOException at all and just let it go. MultiBufferIO in particular throws either ReadException or WriteException so the caller can determine which side of the copy failed. I would declare those explicitly in the `throws`

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
```diff
@@ -69,199 +69,272 @@
...
+            log.error("Failed to read Parquet from InputStream", e);
+            throw new RuntimeException("Error reading Parquet file", e);
+        } finally {
+            try {
+                if (tempFile != null && tempFile.exists()) {
```
pdowler, line 189: So the `tempFile.delete()` here relies on it being opened (line 181-183) above so the content will last as long as it is open? 

Is the underlying file ever closed? Does ParquetReader need a close() method (implement Closeable)? 

While it probably relates to the binary2 PR: does VOTableReader also need to be Closeable for the same reason??

pdowler (changes requested):
I think about half of the io package could be more generally usable, but we can leave it here for now and worry about moving that code to cadc-util later (at which time I'll also compare the RandomAccessSource interface with a similar one used elsewhere).

pdowler (changes requested):
one other thing that I missed

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/ParquetReader.java
aratikakadiya: Updated

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/io/HttpRandomAccessSource.java
```diff
@@ -0,0 +1,154 @@
...
+        // No persistent connection to close
+    }
+
+    private long fetchContentLength(URL url) throws IOException {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        HttpGet get = new HttpGet(url, bos);
+        get.setHeadOnly(true);
+
+        try {
+            get.prepare();
```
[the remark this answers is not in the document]
pdowler: I think that alternate check is dangerous because (iirc):
- if a server accepts the range request, it responds with an indication of "partial content" in the headers
- if it does not accept the range request, it can simple respond with a 200 and the whole stream

At a minimum, that that readAllBytes() could take a considerable amount of time... I think I would not do this at all and only provide random access if the server really intentionally supports it.

cadc-dali/src/main/java/ca/nrc/cadc/dali/util/UUIDFormat.java
```diff
@@ -101,4 +102,29 @@ public String format(UUID t) {
         }
         return t.toString();
     }
+
+    public byte[] uuidToBytes(UUID uuid) {
+        byte[] bytes = new byte[16];
+        long hi = uuid.getMostSignificantBits();
+        long lo = uuid.getLeastSignificantBits();
```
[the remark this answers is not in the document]
pdowler: ok for now. Ideally the UUID <-> byte[] code would be in one place with a round-trip test, but we'll consolidate that later. The code you have for byte[] -> long is quite different from that in HexUtil.toLong(byte[], int) so we'll also chose the best one (that code in HexUtil is reeeaaaally old). But: later.

cadc-dali-parquet/src/main/java/ca/nrc/cadc/dali/tables/parquet/readerhelper/ParquetTableData.java
```diff
@@ -0,0 +1,118 @@
...
+        return new ParquetRowIterator(schema, fields, inputSource);
+    }
+
+    @Override
+    public void close() throws IOException {
+        if (cacheFile != null && cacheFile.exists()) {
+            String fileName = cacheFile.getName();
+            try {
+                if (!cacheFile.delete()) {
+                    throw new java.io.IOException("Failed to delete file: " + fileName);
```
pdowler: don't need fully qualified `java.io.IOException` since it was `import`ed

PR #136, opened 2026-04-23
Cutout bounds fix

at88mph:
# Changes
- [NDimensionalSlicer](https://github.com/opencadc/dal/compare/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/NDimensionalSlicer.java): Fills cornet/span logic when a cutout’s requested bounds extend past the real image, using intersection so lengths and striding are correct instead of capping the difference in a way that can mis-size slices.
- [EnergyCutout](https://github.com/opencadc/dal/compare/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java): Clamps the requested barycentric metre interval to the cube’s spectral extent (linear WCS at channels 1 and N before WCS sky2pix, so e.g. (-Inf, λ) past the high edge still clips to the band instead of no overlap; overlap in pixel space uses a 1…N native span consistent with [FITSCutout.clip](https://github.com/opencadc/dal/compare/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/FITSCutout.java).
- [AxisBoundsFiller](https://github.com/opencadc/dal/compare/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/AxisBoundsFiller.java) (replaces the former “bounds clipper” idea): Puts the clipped 1D interval on the chosen energy/polarization axis and [1, NAXISn] on all other axes (no longer reusing a single “length 60” for every axis).
- [PolarizationCutout](https://github.com/opencadc/dal/compare/cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/PolarizationCutout.java): Returns null when there is no overlap on the Stokes axis (instead of an empty long[]); uses AxisBoundsFiller with per-axis sizes.
- Gradle upgrade for version 8 to allow for JDK 21, and added unit tests

Use case was to fix an Energy cutout where requested bounds exceeded the image cube bounds, but should be clipped.


pdowler:
I do not want to upgrade java version and gradle wrapper in this repo yet... need to evaluate impact on other devs.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/AxisBoundsFiller.java
```diff
@@ -0,0 +1,42 @@
...
+
+
+/**
+ * Fills a flat {@code long[naxis*2]} bounds array: the cut axis uses clipped interval coordinates;
+ * every other axis spans {@code [1, NAXISn]} for that axis (not a single global length).
+ */
+public class AxisBoundsFiller {
+    private static final Logger LOGGER = Logger.getLogger(AxisBoundsFiller.class);
+
+    static long[] fill(final int axes, final long[] clippedBounds, final int clipAxis, final int[] naxisPerAxis) {
```
pdowler: I would probably rename `axes` to `length` since (by usage )it is 2*naxis already -- OR name it `naxis` and do the 2* internally and return a `long[naxis]`... by usage I mean PolarizationCut line 140.

It's certainly confusing and I think use of `naxis` might avoid mistakes.

Is clippedBounds.length == axes and naxisPerAxis.length == axes/2

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java
```diff
@@ -152,39 +162,65 @@ public long[] getBounds(final Interval<Number> bounds) throws NoSuchKeywordExcep
...
+        }
+        return new Interval<>(lo, hi);
+    }
+
+    /**
+     * Wavelength in metres (barycentric) for channels 1 and nchan, using a linear WCS in the native
+     * spectral unit (CUNIT) at the reference pixel (same approximation as a simple 1D grid).
+     * Returns the ordered pair [min, max] in metres.
+     */
+    private double[] spectralWavelengthMetresAtBandEdges(final int energyAxis) {
```
pdowler: This calculation requires using cadc-wcs, pretty much like you see in caom2.git/caom2-compute/..../EnergyUtil.toInterval(SpectralWCS, CoordFunction1D)

At some point I expect to move that code out of caom2, but it isn't practical right now so you probably have to copy and adapt it.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java
```diff
@@ -133,10 +133,20 @@ public long[] getBounds(final Interval<Number> bounds) throws NoSuchKeywordExcep
...
-                    new Interval<>(0.0D, (double) spectralWCSKeywords.getIntValue(
-                            Standard.NAXISn.n(energyAxis).key()));
+            final String ctype = spectralWCSKeywords.getStringValue(Standard.CTYPEn.n(energyAxis).key());
+            final boolean isVelocity = CoordTypeCode.fromCType(ctype).isVelocity();
+            // Intersect user-requested wavelength (m) with the spectral range covered by the data at pixels 1..N
+            // so that extended / infinite physical bounds (e.g. -Inf) map to a finite WCS call instead of
+            // sky2pix endpoints that miss the field of view in pixel space.
+            final Interval<Number> requestMetres = isVelocity ? bounds : clampWavelengthToSpectralFieldOfView(bounds, energyAxis);
+            if (requestMetres == null) {
+                LOGGER.warn("No overlap with spectral extent on data.");
```
pdowler: Can't have warn messages in libraries; this is a normal result that would simply return no content (pixels).

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java
```diff
@@ -133,10 +133,20 @@ public long[] getBounds(final Interval<Number> bounds) throws NoSuchKeywordExcep
...
+                return null;
+            }
+            final int nchan = spectralWCSKeywords.getIntValue(Standard.NAXISn.n(energyAxis).key());
+            final Interval<Double> boundsIntervalPixel = getCutoutPixelInterval(requestMetres, energyAxis, naxis);
+            // FITS pixel indices 1..N; clip() in FITSCutout is 1-based to len inclusive.
+            final Interval<Double> nativePixelsInterval = new Interval<>(1.0D, (double) nchan);
```
pdowler, line 147: Should could be 
```
Interval<Long> nativePixelsInterval
```
if it helps, but iirc WCS pixel coords are also double so it might be ok.

pdowler (changes requested):
Some details that should be looked at carefully.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/AxisBoundsFiller.java
at88mph: Just to keep track, I'm doing one commit per comment.  This stuff tends to be complicated and I hope the tests cover this accurately.  I've covered this in commit `7bc96437644154a3570d56d65de9b798ce8a179c`:
https://github.com/opencadc/dal/pull/136/changes/7bc96437644154a3570d56d65de9b798ce8a179c

pdowler: looks good

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java
at88mph: This is done.  A word of warning, however, it only works until WCSLib 8.4.  After that there was a change to the error codes.  We may have a non-backward compatible version in the near future.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java
at88mph: Done, thanks.

cadc-data-ops-fits/src/main/java/org/opencadc/fits/slice/EnergyCutout.java
at88mph: The spec says floating-point numbers.  Is there an benefit to making it a long?

pdowler (approved):

